PR #44, opened 2013-10-21
Update smtp_util.erl

Gertm:
Tried this with Erlang R16B02 and it failed to compile, this fixes it.


kaos:
And now R18.0 is out.. :)

mworrell:
@Vagabond Ready to drop R15 support?

JustMikey:
@Vagabond R18 is out so if you plan to support only last two major releases, it's time to drop the support for R15.

Vagabond:
I would rather see only supporting the last 2 major releases, so yes.

JustMikey:
@Vagabond The change has been merged from another PR, this can be also closed.

arjan:
Outdated indeed, we moved to support R16 and up only.

PR #68, opened 2014-05-30
use crypto:hmac because crypto:md5_mac is deprecated

matrixise:
md5_mac has been deprecated, use the new function hmac(md5)


JustMikey:
@arjan I believe this can be closed.

arjan:
yes, thanks

PR #71, opened 2014-09-06
Throw permanent_failure when TLS fails due to non-started SSL.

fdevibe:
I spent a bit too much time debugging a temporary_failure, tls_failed error, only to find out that SSL wasn't started. I found this a bit misleading, hence this suggestion to clarify the cause of the failure. Also, I presume permanent_failure would be the correct error type, but I'm not really familiar with the code base, so there may well be better alternatives.


arjan:
Sorry for the enormous delay! :)

PR #72, opened 2014-11-07
convert To address back to string for tokenization

joedevivo:
in `stmp_server_example`, `relay/3` is always called with `binary(), [binary()], binary()` and `string:tokens(<<"ex@ample.com">>, "@")` would always fail.

### DIFF
--- a/src/smtp_server_example.erl
+++ b/src/smtp_server_example.erl
@@ -211,12 +211,12 @@ terminate(Reason, State) ->
 	{ok, Reason, State}.
 
 %%% Internal Functions %%%
-
+-spec relay(binary(), [binary()], binary()) -> ok.
 relay(_, [], _) ->
 	ok;
 relay(From, [To|Rest], Data) ->
 	% relay message to email address
-	[_User, Host] = string:tokens(To, "@"),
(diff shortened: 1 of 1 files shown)


joedevivo:
ping @Vagabond, merge it! :P WWGD?
ganbold

src/smtp_server_example.erl
```diff
@@ -211,12 +211,12 @@ terminate(Reason, State) ->
...
 
 %%% Internal Functions %%%
-
+-spec relay(binary(), [binary()], binary()) -> ok.
 relay(_, [], _) ->
 	ok;
 relay(From, [To|Rest], Data) ->
 	% relay message to email address
-	[_User, Host] = string:tokens(To, "@"),
+	[_User, Host] = string:tokens(binary_to_list(To), "@"),
```
seriyps: Maybe `binary:split(To, <<"@">>)`?

mworrell:
Shouldn't that really parse the email address?
Addresses like `" @ "@example.com` are valid.

arjan:
It's just example code, and the code was clearly broken; this patch just makes it slightly less broken :)

PR #73, opened 2014-11-22
Pass handle_info to the session callback module.

gotthardp:
The gen_smtp_server_session currently ignores unknown messages. The callback module may be interested in receiving for example the 'EXIT' messages. But when an {'EXIT', Pid, Reason} message is received, it is confused with {_SocketType, Socket, Packet} and the gen_smtp_server_session crashes.
I suggest to pass the unhandled messages to the callback module using the handle_info. This is very common in OTP.


Vagabond:
I made this an optional callback, checked for with erlang:function_exported, and merged it.

PR #75, opened 2014-12-14
Set the default encoding to "none" if iconv cannot be found

gotthardp:
The comment in [mimemail.erl:59](https://github.com/vagabond/gen_smtp/blob/master/src/mimemail.erl#L59) suggests the "default encoding is utf-8 if we can find the iconv module", however the default encoding is always "utf-8" even if the iconv is not available.

I suggest to make the dependency on iconv optional by setting the default encoding to "none" if iconv cannot be found.


arjan:
I agree with @mworrell, let's leave it like it is.

PR #77, opened 2015-01-21
Add package.exs for Hex

keichan34:
Hex is a package manager for Elixir (and Erlang) -- making a Hex package makes it a lot easier to use in Elixir projects.

It'd be great if you could publish it to Hex too -- I could do it, but you probably want to make your own releases.

https://hex.pm/docs/publish

### DIFF
--- a/package.exs
+++ b/package.exs
@@ -1,0 +1,24 @@
+defmodule GenSMTP.Mixfile do
+  use Mix.Project
+
+  def project do
+    [app: :gen_smtp,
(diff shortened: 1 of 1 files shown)


mworrell:
I am ok with this, will merge.

I am not sure if @Vagabond wants to publish himself, or that he would prefer you to publish is at this stage.

mworrell:
See #78 for who/when publishes it to Hex.

PR #79, opened 2015-02-12
Fixes crash when username is a binary.

emfa:
The client crashed when doing plain authentication due too the Username beeing a binary.


mworrell:
Thanks!

PR #87, opened 2015-07-09
Added case for handling empty body

KoFish:


seriyps:
Looks good.
Without that case expression `FilteredSize / Size` will fall with `badarith` (division by zero).

mworrell:
Following the logic of the case below the change, wouldn't `7bit` be a more logical encoding?

KoFish:
Thanks, that makes very much sense yes :)

mworrell:
@KoFish thanks!

PR #98, opened 2016-04-07
allow origin  ip/port in tcp/ssl connect options

lazedo:


seriyps:
any notes / explanation?

lazedo:
@seriyps sure, sorry...,  the use case is a server with lot of IP addresses that needs to connect to another relay server that unblocks the firewall only for one of the ip addresses, in this case not the default one.

lazedo:
@seriyps let me know if you need further description.
thanks

seriyps:
But looks like you may just do

```erlang
gen_smtp_client:send(..., [ {sockopts, [ {ip, {1, 2, 3, 4}}, {port, 1234} ]} ])
```

without this patch...

It's not documented, but it's there https://github.com/lazedo/gen_smtp/blob/patch-3/src/gen_smtp_client.erl#L505

And I can't understand how your patch will help with described problem.

lazedo:
`socker:connect` calls `tcp_options` and  `ssl_options` which in turn call  `proplist_merge` that will remove `{ip, {x,x,x,x}}`

lazedo:
to be clear, i use `gen_smtp_client:send(..., [ {sockopts, [ {ip, {1, 2, 3, 4}}, {port, 1234} ]} ])` but it doesn't work without the patch because `ip` and `port` are removed by `tcp_options` or `ssl_options` due to the call to `proplist_merge`

seriyps:
Ah, didn't noticed that proplist_merge drops keys, which are not defined in `DefaultList`.

I'm ok with this change as the simplest one, but maybe we should change the way how `proplist_merge` works? Eg, provide separate list of allowed keys and default values? @Vagabond, @mworrell  ?

lazedo:
@seriyps @Vagabond @mworrell any update here ?

lazedo:
@seriyps @Vagabond @mworrell any update here ?

mworrell:
I am ok with having explicit defaults.

(And yes, we should check that merge code... dropping elements doesn't sound intuitive.)

PR #99, opened 2016-04-14
allow missing "MIME-Version" header

lazedo:
some servers do not set "MIME-Version" header but the content is actually "MIME-Version" : "1.0", so lets try to decode it.

i'm ok with changing `{allow_missing_version, true}` to `{allow_missing_version, false}`. i defaulted to `true` because i think most use cases will benefit from this default

### DIFF
--- a/src/mimemail.erl
+++ b/src/mimemail.erl
@@ -85,9 +85,9 @@ decode(OrigHeaders, Body, Options) ->
 	Encoding = proplists:get_value(encoding, Options, none),
 	%FixedHeaders = fix_headers(Headers),
 	Headers = decode_headers(OrigHeaders, [], Encoding),
-	AllowMissingVersion = proplists:get_value(allow_missing_version, Options, false),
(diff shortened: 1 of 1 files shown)


src/mimemail.erl
```diff
@@ -81,9 +85,13 @@ decode(OrigHeaders, Body, Options) ->
 	Encoding = proplists:get_value(encoding, Options, none),
 	%FixedHeaders = fix_headers(Headers),
 	Headers = decode_headers(OrigHeaders, [], Encoding),
+	AllowMissingVersion = proplists:get_value(allow_missing_version, Options, false),
 	case parse_with_comments(get_header_value(<<"MIME-Version">>, Headers)) of
 		undefined ->
```
seriyps: Maybe move `AllowMissingVersion = proplists:get_value(allow_missing_version, Options, false),` to line after `undefined ->`? So, we can skip this proplists lookup when `MIME-Version` header is defined

lazedo:
@seriyps any update here ?

mworrell:
@lazedo Thanks!

PR #101, opened 2016-04-23
Update smtp_util.erl

JustMikey:
There are no calls to `md5_mac` so I believe we can remove the compiler's `nowarn_deprecated_function`.


arjan:
Thanks!

PR #104, opened 2016-05-16
exclude smtp_rfc822_parse.erl from hex package

tsloughter:
This should fix an issue where the module is being included in the hex package, I noticed this when upgrading to `0.10.0`.


arjan:
Yes, thanks.

PR #109, opened 2016-06-22
Update spec format

benbro:
Old spec format is deprecated on r19


benbro:
I see that there is already PR https://github.com/Vagabond/gen_smtp/pull/106
Any chance it'll be included?

arjan:
#106 is merged, thanks

PR #111, opened 2016-09-30
Convert relay as string if needed

tnt-dev:
The relay should be string.


mworrell:
Makes sense, as we do a `to_string` on other options as well.

Thanks!

PR #112, opened 2016-10-07
Relax MIME-Version header checking

acammack-r7:
Some email senders, notably Yahoo and Amazon SES, send email with a
non-compliant MIME-Version header like:

```
MIME-Version: 1.0
    boundary="----=_Part_505990_122814590.1475767204022"
```

This extra data does not seem important, and discarding it does not
hamper parsing the email.

I have reached out to a few vendors to see what purpose the extra data holds and why it was stuck on the MIME-Version header, but I have yet to receive a response.

### DIFF
--- a/src/mimemail.erl
+++ b/src/mimemail.erl
@@ -239,7 +239,7 @@ convert(To, From, Data) ->
 	Converted.
 
 
-decode_component(Headers, Body, MimeVsn, Options) when MimeVsn =:= <<"1.0">> ->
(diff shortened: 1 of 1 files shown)


Vagabond:
Ugh. I guess we can do this, anyone else have any thoughts?

mworrell:
It is ugly, but then that Email sender is ugly ... guess we need to leak their ugliness into our own code...

@acammack-r7 thanks!

Issue #114, opened 2016-12-19
Unable to send email using office365 smtp servers


minhajuddin:
I want to use office365 smtp endpoint to send out emails. However, for some reason gen_smtp fails with an error:

```erlang
{:error, :retries_exceeded,
 {:temporary_failure, 'smtp.office365.com', :tls_failed}}
```

The code that I am trying to use is below:

```elixir
    :gen_smtp_client.send_blocking({'apisupport@mydomain.com',
                                   ['khaja@mailinator.com'],
                                   ('Subject: Awesome stuff \r\nFrom: apisupport@mydomain.com \r\nTo: khaja@mailinator.com \r\n\r\nThis is the email body'),
                                 },
      [
        {:relay, 'smtp.office365.com'},
        {:port, 587},
        {:hostname,'mydomain.com'},
        {:username, 'apisupport@mydomain.com'},
        {:password, 'password for username'},
        {:tls,:always},
        {:auth, :always},
      ])
```
I have all the required applications started when I check `Applications.started_applications`
```elixir
[{:email_debug, 'email_debug', '0.1.0'},
 {:gen_smtp, 'A generic Erlang SMTP server/client framework', '0.11.0'},
 {:ssl, 'Erlang/OTP SSL application', '8.1'},
 {:public_key, 'Public key infrastructure', '1.3'},
 {:asn1, 'The Erlang ASN1 compiler version 4.0.4', '4.0.4'},
 {:crypto, 'CRYPTO', '3.7.2'}, {:logger, 'logger', '1.3.4'},
 {:mix, 'mix', '1.3.4'}, {:elixir, 'elixir', '1.3.4'},
 {:compiler, 'ERTS  CXC 138 10', '7.0.3'}, {:stdlib, 'ERTS  CXC 138 10', '3.2'},
 {:kernel, 'ERTS  CXC 138 10', '5.1.1'}]
```

I have been trying to fix this for a couple of days without any headway.

I have tested it with ruby and it is working fine which rules out the possibility of auth or port issues.

```ruby
require 'pony'

Pony.mail(
  to: 'khaja@mailinator.com', from: 'apisupport@mydomain.com', subject: "Hello", body: "Hey bud",
  via: :smtp,
  via_options: {
    :address        => 'smtp.office365.com',
    :port           => '587',
    :user_name      => 'apisupport@mydomain.com',
    :enable_starttls_auto => true,
    :password       => 'password',
[...]

PR #115, opened 2016-12-19
Set SSL protocol to tlsv1.2 to avoid using incorrect protocol while using TLS

minhajuddin:
`gen_smtp` sends an SSL request even when `tls` is set to `always` and `ssl` is set to `false`. This causes office365's smtp servers to close the connections before we can upgrade to TLS. This change sets the right version while converting the connection to ssl. See #114 for more info.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -56,6 +56,26 @@ always}` as one of the options. You can specify an alternate port with `{port,
 2525}` (default is 25) or you can indicate that the server is listening for SSL
 connections using `{ssl, true}` (port defaults to 465 with this option).
 
+Options
+=======
+    send(Email, Options)
+    send(Email, Options, Callback)
+    send_blocking(Email, Options)
+
+The `send` method variants `send/2, send/3, send_blocking/2` take an `Options` argument.
+`Options` must be a proplist with the following valid values:
+
+  * **relay** the smtp relay, e.g. `"smtp.gmail.com"`
+  * **username** the username of the smtp relay e.g. `"me@gmail.com"`
+  * **password** the password of the smtp relay e.g. `"mypassword"`
+  * **auth** whether the smtp server needs authentication, valid values are `if_available` and `always`, Defaults to `if_available`. If your smtp relay requires authentication set it to `always`
+  * **ssl** whether to connect on 465 in ssl mode, Defaults to `false`
+  * **tls** valid values are `always`, `never`, `if_available`. Most modern smtp relays use tls, so set this to `always`, Defaults to `if_available`
+  * **tls_options** used in ssl:connect, More info at http://erlang.org/doc/man/ssl.html , Defaults to [{versions , ['tlsv1.1', 'tlsv1.2']}], This is merged with options listed at: https://github.com/Vagabond/gen_smtp/blob/master/src/socket.erl#L46 . Any options not present in this list will be ignored.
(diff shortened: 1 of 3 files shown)


Vagabond:
Yeah, I think this is on the right track but we should expose a generic way to pass TLS/SSL options, and not simply hardcode this one tunable.

We can probably also kill `{ssl_imp, new}`, I'm pretty sure that option is long dead.

minhajuddin:
@Vagabond Thanks for responding, Should we change this to allow users to pass it via `gen_smtp_client:send` as part of the `Options` ? we could add a new option named `ssl_tls_options` which would allow a user to pass a list of ssl_tls options.

minhajuddin:
@Vagabond I have changed the code based on your comments. Have a look.

Vagabond:
This looks pretty good, just a few suggestions:

* Call it just `tls_options`, SSL is pretty dead as a protocol.
* Don't change the default protocol version, we might break a lot of existing usage. Alternatively, use a slightly less strict default (1.1 and 1.2, maybe?)
* Documentation on the new option.

minhajuddin:
@Vagabond Thanks for the feedback, Committed changes based on your feedback.

README.markdown
```diff
@@ -56,6 +56,26 @@ always}` as one of the options. You can specify an alternate port with `{port,
...
+The `send` method variants `send/2, send/3, send_blocking/2` take an `Options` argument.
+`Options` must be a proplist with the following valid values:
+
+  * **relay** the smtp relay, e.g. `"smtp.gmail.com"`
+  * **username** the username of the smtp relay e.g. `"me@gmail.com"`
+  * **password** the password of the smtp relay e.g. `"mypassword"`
+  * **auth** whether the smtp server needs authentication, valid values are `if_available` and `always`, Defaults to `if_available`. If your smtp relay requires authentication set it to `always`
+  * **ssl** whether to connect on 465 in ssl mode, Defaults to `false`
+  * **tls** valid values are `always`, `never`, `if_available`. Most modern smtp relays use tls, so set this to `always`, Defaults to `if_available`
+  * **tls_options** used in ssl:connect, More info at http://erlang.org/doc/man/ssl.html , Defaults to [{versions , ['tlsv1.1', 'tlsv1.2']}], This is merged with options listed at: https://github.com/Vagabond/gen_smtp/blob/master/src/socket.erl#L46 . Any options not present in this list will be ignored.
```
Vagabond: Maybe we should not filter the list anymore, and just allow any option to be passed to the merge? I'm not sure why I ever did that in the first place.

Vagabond:
This looks really, really nice. I think we should probably support TLS 1.0 by default as well because it isn't sunsetted until 2018 (ugh) and we don't want to break anyone using the existing defaults?

README.markdown
minhajuddin: From git blame it looks like it was done to avoid incorrect options from being passed: https://github.com/Vagabond/gen_smtp/commit/4b70b227ab5eb0a69d9c9ec27817c9c770606cd9

minhajuddin:
@Vagabond Added tlsv1 to the default list

Vagabond:
Yeah, but invalid options will break things, silently failing and possibly ignoring new options seems like a bad way to handle things. I think we should revisit that logic, but I guess we are getting a bit out of scope for your PR, sorry.

PR #117, opened 2017-01-24
Enable sending multiple emails in a single SMTP client session

Vagabond:
This allows you to supply a list of email tuples to the smtp_client functions. Other fixes:

* Fixed a bunch of broken tests
* Fallback to A records if there's no MX records (eg smtp.gmail.com)
* Add a trace_fun option you can use to print/log the smtp session progress (eg `fun io:format/2`)

This PR has some problems, however. Firstly, right now, it breaks backwards compatability with the send_blocking API, it will currently return a list of results, not just one.

Secondly, and much worse, the partial failure modes for this become *extremely* tricky. What happens if you're sending 5 emails and the third one is invalid, right now it will retry the *entire* session, so the first 2 emails will send multiple times, and the last emails will never be attempted at all. This is not trivial to fix, as far as I can tell, suggestions welcome.


seriyps:
It would be nice to add some update to README about this new multi-email feature.
Also, some tests.

skosch:
Wow. Thank you, thank you!

Breaking the API would be problematic for other libraries (e.g. [kamilc/mailman](https://github.com/kamilc/mailman)) so it's probably not a good idea to mess with that. No need to get fancy though, I think this could totally be a separate function.

Ultimately, when you're sending multiple emails, retries have to be handled on a per-email basis. So far my user-land approach has been to tag all outgoing emails with `retries: 10` and to recursively re-attempt the delivery of all those with `retries >= 0` – see below, Elixir, sorry. (I then merrily open a whole bunch of connections in parallel, if some don't make it, so be it :smiling_imp:)

I don't see a way around a similar kind of logic in this case, except for the added difficulty that different errors must trigger different things, such as
* on timeout: decrement `retries`, then reconnect SMTP
* on permanent failure: set `retries` = 0 and return the error

IMHO this will only work if establishing the SMTP connection is decoupled from sending a single email, but you [already knew that anyway](https://github.com/Vagabond/gen_smtp/issues/116#issuecomment-270968879) :)

```elixir
  # to deliver all, call deliver_all_gen_smtp(envelopes)

  defp deliver_all_gen_smtp(_retryables, _nonretryables \\ [])
  defp deliver_all_gen_smtp([], nonretryables) do
    {:ok, "finished", nonretryables}
  end
  defp deliver_all_gen_smtp(retryables, nonretryables) do
    delivery_results = 
      retryables
      |> Task.async_stream(&deliver_one(&1), max_concurrency: @max_concurrency, timeout: @timeout)
      |> Enum.to_list  # async_stream *should* return {:ok, ..} tuples; we remove the :ok
      |> Enum.filter_map(&match?({:ok, _}, &1), fn {:ok, result} -> result end)

    {new_nonretryables, retryables} =
      delivery_results
      |> Enum.partition(fn envelope ->
        envelope.delivered or envelope.retries <= 0
      end)

    # Try again with those that timed out previously
    deliver_all_gen_smtp(retryables, new_nonretryables ++ nonretryables)
  end

  defp deliver_one(envelope) do
    case gen_smtp_deliver(envelope) do
      {:ok, delivered_message} ->
        Logger.debug("Successfully delivered message to #{envelope.envelope_to}.")
        log_delivery_date(envelope, delivered_message)
        %{envelope | delivered: true}

      {:error, :no_more_hosts, {:permanent_failure, server_address, error_message}} ->
        Logger.warn("Failed to deliver to #{envelope.envelope_to} due to #{error_message}. Not trying again.")
        log_delivery_error(envelope, server_address, error_message)
        %{envelope | delivered: false, retries: 0}

      {:error, :retries_exceeded, {:network_failure, _server_address, {:error, :timeout}}} ->
        Logger.warn("Timed out while trying to deliver to #{envelope.envelope_to}. Trying again #{envelope.retries - 1} more times.")
        %{envelope | delivered: false, retries: envelope.retries - 1}

      err -> 
        log_delivery_error(envelope, "", inspect(err))
        Logger.warn("Email could not be sent: #{inspect(err)}")
        nil
    end
  end
```

skosch:
How about an approach like this (my apologies for the atrocious pseudocode, it's 6AM and I haven't even had coffee)
```
do_smtp_session([{email, retries, email_index} | retryable_email_tail], nonretryable_emails, {socket, socket_status}) ->
  if socket_status == socket_dead:
    socket = reconnect()
  
  if retries > 0:
    error = try_sending_it(email)
    case result do
      timeout -> do_smtp_session([{email, retries - 1, email_index} | retryable_email_tail], nonretryable_emails, {socket, socket_dead})
      permanent_failure -> do_smtp_session(retryable_email_tail, [{email, error, email_index} | nonretryable_emails], {socket, socket_alive}) # or is it dead here too?
      receipt -> do_smtp_session(retryable_email_tail, [{email, receipt, email_index} | nonretryable_emails], {socket, socket_alive})
    end
  else
    do_smtp_session(retryable_email_tail, [{email, no_more_retries, email_index} | nonretryable_emails], {socket, socket_status})
  end.


do_smtp_session([], nonretryable_emails_with_results, _) ->
  sort_by_email_index(nonretryable_emails_with_results).
```

seriyps:
Don't we need to call `RSET` after each delivery? https://tools.ietf.org/html/rfc5321#section-4.1.1.5

skosch:
Entirely possible; my working knowledge of SMTP is minimal. Would that thwart the above implementation?

Vagabond:
Yeah, I honestly don't remember anymore. It has been so long since I read the relevant RFCs.

I think we can do it in a backwards compatible way. I will ponder it more, but an alternative form where you just get given a socket with the SMTP session established (and functions to work with it), would also be useful.

I'm not sure I'll have time to do much more on this for a while, but maybe this is a useful starting point for someone else?

skosch:
>  an alternative form where you just get given a socket with the SMTP session established (and functions to work with it)

Yeah, I think that might be enough for now – if you just provide a friendly API to the key functions in `gen_smtp_client` (i.e. sending and re-establishing the session if necessary) we could build our own logic out of those.

Vagabond:
Ok, I will look into that instead, although I still don't have a lot of time for this.

skosch:
No rush @Vagabond, I didn't mean to bully you into this :bowing_man: 

As I mentioned, I've already got a working system, so there's no time pressure, it'd just be nice to have.

Vagabond:
```1> S = gen_smtp_client:open([{relay, "smtp.gmail.com"}, {username, "me@gmail.com"}, {port, 587}, {password, "password"}, {trace_fun, fun io:format/2}]).
connected to 74.125.28.108; banner was  smtp.gmail.com ESMTP 187sm9109060pgc.46 - gsmtp

Extensions are [{<<"SIZE">>,<<"35882577">>},
                {<<"8BITMIME">>,true},
                {<<"STARTTLS">>,true},
                {<<"ENHANCEDSTATUSCODES">>,true},
                {<<"PIPELINING">>,true},
                {<<"CHUNKING">>,true},
                {<<"SMTPUTF8">>,true}]
Starting TLS
TLS started
Extensions are [{<<"SIZE">>,<<"35882577">>},
                {<<"8BITMIME">>,true},
                {<<"AUTH">>,
                 <<"LOGIN PLAIN XOAUTH2 PLAIN-CLIENTTOKEN OAUTHBEARER XOAUTH">>},
                {<<"ENHANCEDSTATUSCODES">>,true},
                {<<"PIPELINING">>,true},
                {<<"CHUNKING">>,true},
                {<<"SMTPUTF8">>,true}]
Auth types: <<"LOGIN PLAIN XOAUTH2 PLAIN-CLIENTTOKEN OAUTHBEARER XOAUTH">>
Fixed types: ["LOGIN","PLAIN","XOAUTH2","PLAIN-CLIENTTOKEN","OAUTHBEARER",
              "XOAUTH"]
available authentication types, in order of preference: ["LOGIN","PLAIN"]
username prompt
password prompt
authentication accepted
Authentication status is true
{{sslsocket,{gen_tcp,#Port<0.39010>,tls_connection,
                     undefined},
            <0.133.0>},
 [{<<"SIZE">>,<<"35882577">>},
  {<<"8BITMIME">>,true},
  {<<"STARTTLS">>,true},
  {<<"ENHANCEDSTATUSCODES">>,true},
  {<<"PIPELINING">>,true},
  {<<"CHUNKING">>,true},
  {<<"SMTPUTF8">>,true}],
 [{auth,if_available},
  {hostname,"system76"},
  {password,"patch26"},
  {port,587},
  {relay,"smtp.gmail.com"},
  {retries,1},
  {ssl,false},
  {tls,if_available},
  {tls_options,[{versions,[tlsv1,'tlsv1.1','tlsv1.2']}]},
  {trace_fun,#Fun<io.format.2>},
  {username,"tetrachloride@gmail.com"}]}
2> gen_smtp_client:deliver(S, {"whatever@test.com", ["andrew@hijacked.us"],  "Subject: testing\r\nFrom: Andrew Thompson \r\nTo: Some Dude \r\n\r\nThis is the email body"}).
<<"2.0.0 OK 1495949758 187sm9109060pgc.46 - gsmtp\r\n">>
```

skosch:
:open_mouth:

Vagabond:
if this looks like a better API, I can try to clean it up and merge it. Feedback appreciated.

skosch:
Sorry, haven't looked at it closely yet, but your example looks great! Will this break libraries relying on your code? That would be my only concern.

Vagabond:
The API should be backwards compatible, or that's a bug.

skosch:
`send` and `send_blocking` with a list of emails will no longer work, right? Although I don't think that's a problem, not for mailman at least. I'd say go ahead :)

Vagabond:
Yeah, I removed the list form, it didn't seem like a workable API.

Vagabond:
I've rebased this and it seems in decent shape now, anyone care to review?

cc @mworrell @arjan

Vagabond:
I added some edoc and dialyzer stuff, any objections to this merging soon?

src/gen_smtp_client.erl
```diff
@@ -121,7 +123,46 @@ send_it_nonblock(Email, Options, Callback) ->
 			{ok, Receipt}
 	end.
 
--spec send_it(Email :: {string() | binary(), [string() | binary(), ...], string() | binary() | function()}, Options :: list()) -> binary() | {'error', any(), any()}.
+-opaque smtp_client_socket() :: {socket:socket(), list(), list()}.
```
mworrell: Maybe move this definition closer to the export?
Was also wondering if we shouldn't use a record here, for future changes?

src/gen_smtp_client.erl
```diff
@@ -121,7 +123,46 @@ send_it_nonblock(Email, Options, Callback) ->
...
+									[{0, RelayDomain}]; % maybe we're supposed to relay to a host directly
+								_ ->
+									MXRecords
+							end,
+			try_smtp_sessions(Hosts, NewOptions, []);
+		{error, Reason} ->
+			{error, bad_option, Reason}
+	end.
+
+-spec deliver(Socket :: smtp_client_socket(), Email :: email()) -> Receipt :: binary() | {error, any()}.
```
mworrell: Shall we make the return value `{ok, binary()}`, for easier pattern matching?

The same for the `open/1` function: `{ok, smtp_client_socket()}`

src/gen_smtp_client.erl
```diff
@@ -47,12 +47,14 @@
 -include_lib("eunit/include/eunit.hrl").
 -compile(export_all).
 -else.
--export([send/2, send/3, send_blocking/2]).
+-export([send/2, send/3, send_blocking/2, open/1, deliver/2]).
```
mworrell: No `close/1` function?
How do we clean up an opened connection?
(I assume I can use `deliver/1` multiple times on an open connection)

mworrell:
I like the idea of opening a connection and then using *deliver* to actually send emails across.

The `trace` function is also a great addition.

Vagabond:
Updates made. I don't think we need a record yet, but it is an opaque type, so we can change the representation at any time.

mworrell:
All looks good to me! 👍 for merging this.

PR #120, opened 2017-03-02
close socket properly during session init when connection is not exist anymore

icehess:
We observed when a connection is opened and closed in a short amount of time before even `gen_smtp_server_session` is started, the call to `socket:peername/1` causes crash in  `gen_smtp_server_session` and the socket remains open.

If this kind of connections happen periodically (e.g. during a Haproxy health check) , the node eventually would run out of file descriptors.

This tries to handle the call to `socket:peername/1` and will close the socket if the connection is closed during initializing of a session.

### DIFF
--- a/src/gen_smtp_server_session.erl
+++ b/src/gen_smtp_server_session.erl
@@ -113,8 +113,16 @@ start(Socket, Module, Options) ->
 %% @private
 -spec init(Args :: list()) -> {'ok', #state{}, ?TIMEOUT} | {'stop', any()} | 'ignore'.
 init([Socket, Module, Options]) ->
-	{ok, {PeerName, _Port}} = socket:peername(Socket),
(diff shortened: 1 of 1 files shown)


src/gen_smtp_server_session.erl
```diff
@@ -113,8 +113,16 @@ start(Socket, Module, Options) ->
...
+		{ok, {IPaddr, _Port}} -> IPaddr;
+		{'error', _} -> 'error'
+	end,
+	case PeerName =/= 'error'
+		andalso Module:init(proplists:get_value(hostname, Options, smtp_util:guess_FQDN()), proplists:get_value(sessioncount, Options, 0), PeerName, proplists:get_value(callbackoptions, Options, []))
+	of
+		false ->
+			socket:close(Socket),
+			ignore;
 		{ok, Banner, CallbackState} ->
```
mworrell: The 'andalso' suggests a boolean check, I would prefer to have the PeerName check and the init return handled separately.  In this case you can have an `init_peername({error, _}, ...)` that closes the socket and a `init_peername({ok, ...}, ...)` that handles the init.

mworrell:
Merging anyway, as the change is important enough and works well.

PR #127, opened 2017-06-06
Fix for connecting to Hotmail/Outlook servers

mworrell:
On OTP 19 gen_smtp has problems connecting to Hotmail/Outlook (and other Microsoft servers).

The server closes the connection during the handshake.
This problem seems to very similar to the issue below:

http://erlang.org/pipermail/erlang-bugs/2016-September/005195.html

Setting the default TLS version to TLSv1.2 fixes this problem.

As v1.2 has been introduced in 2008, it seems quite safe to me to set this as the default.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,4 +11,5 @@ src/smtp_rfc822_parse.erl
 deps/
 _build
 .rebar
+.rebar3
 compile_commands.json
--- a/src/gen_smtp_client.erl
+++ b/src/gen_smtp_client.erl
@@ -28,7 +28,7 @@
 -define(DEFAULT_OPTIONS, [
 		{ssl, false}, % whether to connect on 465 in ssl mode
 		{tls, if_available}, % always, never, if_available
-		{tls_options, [{versions, ['tlsv1', 'tlsv1.1', 'tlsv1.2']}]}, % used in ssl:connect, http://erlang.org/doc/man/ssl.html
+		{tls_options, [{versions, ['tlsv1.2']}]}, % used in ssl:connect, http://erlang.org/doc/man/ssl.html
 		{auth, if_available},
 		{hostname, smtp_util:guess_FQDN()},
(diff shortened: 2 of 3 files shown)


mworrell:
Testing this in production we found a problem with upcmail.nl, mail to that server with only TLSv1.2 returned:

```
{error, no_more_hosts, {permanent_failure, _Host, <<105,103,110,32,82,111,111,116,32, _/binary>>}}
```

Where the binary is *very* long (and indeed binary).  Something seems to go wrong with the TLS upgrade to `upcmail.nl`. The MX server is `mx0.upcmail.net`

Vagabond:
certainly we should no longer even try sslv3, I'm on the fence about tls 1.0. We probably still need to support 1.1 and 1.2.

mworrell:
There clearly are still some older servers around.
But then, as they only use (very) old TLS, they probably also accept plain text.
Which could then be a valid fallback.

Vagabond:
TLS uptake does not keep pace with when the spec is published. TLS 1.0 is still fairly widespread, but trying to use sslv3 will actually result in the connection being dropped, increasingly.

Vagabond:
#115 has some related details here.

mworrell:
If we want to use TLSv1 and v1.1, then we need a way to handle Microsoft servers, as they only accept our handshake if we specify v1.2.

It seems a fix is coming up in OTP 20:

https://bugs.erlang.org/browse/ERL-206?jql=project%20%3D%20ERL%20AND%20component%20%3D%20ssl%20ORDER%20BY%20updated%20DESC%2C%20priority%20DESC%2C%20created%20ASC

mworrell:
A fix should be in for OTP 20, somewhere later this month.

Instead of this merge-request I will make the detection of invalid handshakes more robust so that we have a better fallback to non-lts connections.

Closing this for now.

PR #129, opened 2017-07-12
Add passthrough for x-binaryenc charset

acammack-r7:
Not an IANA standard and not used for text, but does show up in emails
sent from some clients. Since it is the same as no encoding, we pass it
through unmodified.


mworrell:
I see that some Google software uses this content-type, so might be actually in *the wild*.

Merging this.

Thanks!

PR #130, opened 2017-07-12
Add check that port is an integer

### LINKED ISSUE #128: No error when port is not an integer
If the `port ` option passed as part of the `Options` proplist to one of the functions in `gen_smtp_client` is a binary and not an integer, gen_smtp silently falls back to using one of the default ports. I think it should instead return an error just like when no relay has been specified.
What do you think? Would you accept a pull request for this?

wmnnd:
Fixes #128

### DIFF
--- a/src/gen_smtp_client.erl
+++ b/src/gen_smtp_client.erl
@@ -651,7 +651,7 @@ quit(Socket) ->
 
 % TODO - more checking
 check_options(Options) ->
-	CheckedOptions = [relay, auth],
+	CheckedOptions = [relay, port, auth],
 	lists:foldl(fun(Option, State) ->
 		Value = proplists:get_value(Option, Options),
 		case State of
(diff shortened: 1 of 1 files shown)


src/gen_smtp_client.erl
```diff
@@ -651,23 +651,30 @@ quit(Socket) ->
...
+			Other -> Other
+		end
+	end, ok, CheckedOptions).
+
+check_option({relay, undefined}, _Options) -> {error, no_relay};
+check_option({relay, _}, _Options) -> ok;
+check_option({port, undefined}, _Options) -> ok;
+check_option({port, Port}, _Options) when is_integer(Port) -> ok;
+check_option({port, _}, _Options) -> {error, invalid_port};
+check_option({auth, Auth}, Options) when Auth =:= always ->
```
mworrell: Maybe the following is clearer?

```Erlang
check_option({auth, always}, Options) ->
```

wmnnd:
Thanks for the feedback, @mworrell! I have implemented your suggestion.

mworrell:
Looks good.

Will merge, thank you for your changes!

Issue #95, opened 2016-02-29
Remove deprecated eiconv dependency


edevil:
eiconv has been deprecated as per https://github.com/processone/eiconv

It would be nice to switch to the supported iconv lib: https://github.com/processone/iconv/

mworrell:
`gen_smtp` is referring to the similarly named https://github.com/zotonic/eiconv, not to the ProcessOne eiconv

mworrell:
We are using the Zotonic `eiconv` because the other NIFs block the scheduler on large input texts. The Zotonic eiconv solves this by chopping the input into smaller chunks and converting those chunks one by one.

A quick inspection of the ProcessOne `iconv` shows that they don't have a routine to chunk the input, so I expect it will block the Erlang scheduler on large inputs.

I suggest to either stick with Zotonic `eiconv` or see if you can get the chunking merged into the ProcessOne version.

seriyps:
Can be closed?

mworrell:
I think it can be closed, maybe revisit when the ProcessOne iconv lib supports chunks.

PR #132, opened 2017-08-21
Use new iconv:convert/3 API instead of iconv:open/2 / iconv:conv/2 / iconv:close/1

MonsieurV:
Hi,

I've tried to migrate from eiconv to iconv #95.

It works well... until an iconv conversion fails.

All tests pass except two:
* `permissive malformed folded multibyte header decoder` causes a `segfault` in iconv module, which brings down the Erlang VM;
* `invalid character sequence handling` fails to convert the Copyright logo, but do not raise an error as expected. (If I understand it well)

Because of the segfault issue, it still need some work so we keep the server reliable.

For that, we certainly want to make a PR on [processone/iconv](https://github.com/processone/iconv) for better error handling and reporting. processone/iconv#6 could be a good start.

Yoan

PS. Review welcome, I'm an Erlang newbie!


mworrell:
See my comments on #95 

> We are using the Zotonic `eiconv` because the other NIFs block the scheduler on large input texts. The Zotonic eiconv solves this by chopping the input into smaller chunks and converting those chunks one by one.
> 
> A quick inspection of the ProcessOne `iconv` shows that they don't have a routine to chunk the input, so I expect it will block the Erlang scheduler on large inputs.
> 
> I suggest to either stick with Zotonic `eiconv` or see if you can get the chunking merged into the ProcessOne version.

MonsieurV:
Based on what you said (and reviewing the two iconv modules), ([Zotonic/eiconv](https://github.com/zotonic/eiconv) seems just more complete.

So finally I've just changed the call to the new iconv API (`convert/3`), while sticking to Zotonic `eiconv`.

Maybe it would be interesting to also support [processone/eiconv](https://github.com/processone/iconv). There are two issues for that:
* the function we want is named `iconv:convert/3` and not `eiconv:convert/3`
* ProcessOne `iconv:convert/3` returns directly the converted text and not a tuple `{ok, Result}`

One way could be to mimic the ProcessOne `convert/3` API in Zotonic [iconv](https://github.com/zotonic/eiconv/blob/debcb9f4e3aa2026b0cd03843170ea9cd5662b85/src/iconv.erl#L15). It will add a new API without changing any of the existing API.

The other way is to submit an API change to ProcessOne to at least returns a tuple `{ok, Result}`, but this will break their API.

What do you think @mworrell?

Actually the first reason why for me to propose this PR was that I was unable to load an `iconv` module properly and make the gen_smtp parsing works. This was just ignorance on my part.

I added documentation to make the process clearer, for guys that are news to Erlang like me. Also that makes the dependency clearly stated for those that want to use the parsing.

Thanks @mworrell

MonsieurV:
[emqplus/erlang-iconv](https://github.com/emqplus/erlang-iconv/blob/master/src/iconv.erl#L92) also has an `iconv:convert/3` API that returns directly the converted text.

So we way follow this API to support the most iconv wrappers possible.

MonsieurV:
There has been a PR on zotonic/eiconv/pull/2 to support the `iconv:convert/3` interface.

This PR can now be merged!

It allows people to use one of the following `iconv` Erlang wrapper for decoding mails:
* [zotonic/eiconv](https://github.com/zotonic/eiconv)
* [emqplus/erlang-iconv](https://github.com/emqplus/erlang-iconv)
* [processone/iconv](https://github.com/processone/iconv)

The process for setup the server the first time should also be easier (i.e. more documented) for beginners.

MonsieurV:
Or maybe @arjan,  can you check it? :)

arjan:
Thanks @MonsieurV I'll have a look later today.

MonsieurV:
@arjan I forgot this one! Can we have a look?

MonsieurV:
@arjan It would be super cool to check the PR and publish a new version on hex.pm

I rely on `gen_smtp` for one of [my library](https://github.com/YtoTech/geo_sensors_gateway) and I currently can't publish it on Hex (all the dependencies must themselves be on hex.pm)

Can I help somehow?

MonsieurV:
Hey!

Sorry for pushing, but can we have any chance to merge and publish this change to Hex?

Vagabond:
Do we have a plan for this? @arjan @mworrell

arjan:
Will be merged through PR #177

PR #136, opened 2017-10-24
Replace crypto:rand_uniform/2 with rand:uniform/1

kyle-neal:
Simple change to make gen_smtp compliant with OTP 20+

### DIFF
--- a/src/smtp_util.erl
+++ b/src/smtp_util.erl
@@ -68,7 +68,7 @@ compute_cram_digest(Key, Data) ->
 %% @doc Generate a seed string for CRAM.
 -spec get_cram_string(Hostname :: string()) -> string().
 get_cram_string(Hostname) ->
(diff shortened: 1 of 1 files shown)


seriyps:
But it makes it incompatible with OTP < 18.

seriyps:
Plus there is no such function as `rand:uniform/2`. Only `rand:uniform/1` http://erlang.org/doc/man/rand.html#uniform-1.

You may try to use `crypto` module: http://erlang.org/doc/man/crypto.html#rand_uniform-2

mworrell:
@Vagabond @arjan would you be ok to drop support for R16B and OTP-17?

I am ok with that, as OTP 21 has been released already.

mworrell:
Closing this and not merging, as the function `rand:uniform/2` does not exist.

PR #144, opened 2018-03-18
change Obsolete API function to_upper to uppercase

sunchess:
In elixir function string:to_upper doesn't work
```
iex(1)> :string.to_upper("test")
** (FunctionClauseError) no function clause matching in :string.to_upper/1

    The following arguments were given to :string.to_upper/1:

        # 1
        "test"

    (stdlib) string.erl:1810: :string.to_upper/1

```
but string:uppercase works
```
iex(2)> :string.uppercase("test")
"TEST"
```

### DIFF
--- a/src/gen_smtp_server_session.erl
+++ b/src/gen_smtp_server_session.erl
@@ -704,8 +704,8 @@ parse_encoded_address(<<H, Tail/binary>>, Acc, Quotes) ->
 
 -spec has_extension(Extensions :: [{string(), string()}], Extension :: string()) -> {'true', string()} | 'false'.
 has_extension(Exts, Ext) ->
-	Extension = string:to_upper(Ext),
-	Extensions = [{string:to_upper(X), Y} || {X, Y} <- Exts],
+	Extension = string:uppercase(Ext),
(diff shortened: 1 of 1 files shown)


bryanhuntesl:
@sunchess - that's because gen_smtp is operating on character lists as opposed to strings. Strings in Elixir are binaries, 'strings' in Erlang are lists. Change the quoting and the function works correctly. The function `:string.uppercase/1` also works with binaries. 

Example (erlang lists)

```
iex(26)> :string.to_upper('foo')
'FOO'
iex(27)> :string.uppercase('foo')
'FOO'
```

Example (erlang binaries)
```
iex(28)> :string.uppercase(<<"foo">>)
"FOO"
iex(29)> :string.to_upper(<<"foo">>)
** (FunctionClauseError) no function clause matching in :string.to_upper/1

    The following arguments were given to :string.to_upper/1:

        # 1
        "foo"

    (stdlib) string.erl:2096: :string.to_upper/1
```

Yeah, Erlang APIs are uh, interesting.

sunchess:
Hey, @bryanhuntesl seems yes! I just copied
` Extensions ++ [{"AUTH", "PLAIN LOGIN CRAM-MD5"}]; `

PR #147, opened 2018-06-21
Fix problem retrying when the socket is closed during starttls. Fixes #146

### LINKED ISSUE #146: When the socket is closed during ssl handshake no retry is made to send mail without ssl
I was debugging a connection problem to outlook servers. 

I found that OTP 19.3 can't do a TLS upgrade to outlook servers. I don't know why. Connecting with OTP 18 works fine.

When this happens `socket:to_ssl_client` returns `{error, closed}`. This is not handled correctly. The underlying socket is closed, which leads to a crash later on. When this happens gen_smtp won't retry to send the mail without using TLS.

Do you want a PR with a fix for this problem?

mmzeeman:
Handle the situation where the smtp server closes the socket during starttls.

Fixes a typo during the retry. The options are changed, they where matched against the old options.

### DIFF
--- a/src/gen_smtp_client.erl
+++ b/src/gen_smtp_client.erl
@@ -187,12 +187,12 @@ send_it(Email, Options) ->
 			MXRecords
 	end,
 	case try_smtp_sessions(Hosts, Options, []) of
-		{Socket, Extensions, Options} ->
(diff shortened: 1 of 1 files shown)


mworrell:
Thanks!

PR #149, opened 2018-06-27
fix io_lib:format match error on Erlang 21.0

### LINKED ISSUE #148: Error when sending some emails with transfer encoding = "quoted-printable" on Erlang OTP 21.0
[This line](https://github.com/Vagabond/gen_smtp/blob/2d347f81dc354c3ee78ccbc22efabc5c08260552/src/mimemail.erl#L905) fails with match error because on 20.3 `io_lib:format('~2.16.0B', [10]).` returns `[["0",65]]`, and on 21.0 same line return `"0A"` (without second wrapping list).

azhi:
`io_lib:format('~2.16.0B', [10])` returned `[["0", 65]]` on 20.3, but `"0A"`
on 21.0

Fixes #148

### DIFF
--- a/src/mimemail.erl
+++ b/src/mimemail.erl
@@ -902,7 +902,7 @@ encode_quoted_printable(<<H, $\r, $\n, T/binary>>, Acc, _L) when H == $\s; H == 
 encode_quoted_printable(<<H, T/binary>>, Acc, L) when H == $\s; H == $\t ->
 	encode_quoted_printable(T, [H | Acc], L+1);
 encode_quoted_printable(<<H, T/binary>>, Acc, L) ->
-	[[A, B]] = io_lib:format("~2.16.0B", [H]),
+	[A, B] = lists:flatten(io_lib:format("~2.16.0B", [H])),
 	encode_quoted_printable(T, [B, A, $= | Acc], L+3).
 
 get_default_encoding() ->


Vagabond:
Can you not change whitespace at the same time? Otherwise this looks good (although we could also use lists:flatten() instead of list_to_binary()).

mworrell:
Agree, please next time only change the code and not the whitespace (though I understand your editor might have done it automatically)

@Vagabond merge as-is?

azhi:
Updated: use `lists:flatten` instead of converting to binary, and remove all whitespace fixes (yes, I have whitespace autoformatting enabled, that's generally a good idea :))

mworrell:
Thanks!

joefreeman:
I think we might need the same fix for line [900](https://github.com/Vagabond/gen_smtp/pull/149/files#diff-7878b2f6d5a7b3b1d17f539d33c55d66L900) as well?

azhi:
My bad, missed it. I'll create another PR.

JustMikey:
Apparently this bug start to cause crashes when moved to OTP 21. Any chance we can bump up the package version?

Vagabond:
I tagged 0.13, but I can't push to hex. @arjan can you add me as a hex admin for the project?

mworrell:
It seems @keichan34 has also ownership on Hex

Vagabond:
Yeah, it looks like you found me, thanks!

mworrell:
@keichan34 always good to have a backup for if @Vagabond or @arjan aren't available

PR #153, opened 2018-07-16
Remove overly-rigid pattern-matching in guess_FQDN

alanj853:
In the event that there is no DNS record for a particular hostname,
we simply handle the error and return a charlist instead of crashing.
We also handle all other errors to avoid crashing.


alanj853:
Created accidentally. Closing now.

PR #154, opened 2018-07-16
Fix for https://github.com/Vagabond/gen_smtp/issues/152

alanj853:


Vagabond:
The new return types you have aren't handled in the places this function is called.

alanj853:
@Vagabond , yes apologies. Returning tuples requires much more time and testing that I can afford to give right now. So I have compromised, and am once again returning a character list, but I have the exact error inserted into it.

Vagabond:
But if you do that, it looks like the DNS resolution suceeded. If a function returns more than one kind of 'thing' you need to be able to distinguish those kinds. This is why the `{ok, Result} | {error, Reason}` tuple patterns are so prevalent.

arjan:
Closing this PR due to inactivity.

PR #156, opened 2018-08-07
warning: Update code to fix warnings emitted by the compiler

tomciopp:
Updates code to fix compiler warnings. rand:unform/1 has a range from 1 to x instead of 0, but in this case I don't believe that matters. ssl:ssl_accept has been deprecated in OTP 21 and they recommend using handshake instead. 

http://erlang.org/doc/man/ssl.html#ssl_accept-3

### DIFF
--- a/src/smtp_util.erl
+++ b/src/smtp_util.erl
@@ -35,7 +35,7 @@ mxlookup(Domain) ->
 	case whereis(inet_db) of
 		P when is_pid(P) ->
 			ok;
-		_ -> 
+		_ ->
 			inet_db:start()
 	end,
 	case lists:keyfind(nameserver, 1, inet_db:get_rc()) of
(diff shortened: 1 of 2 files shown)


Vagabond:
ssl:handshake was only added in OTP21, we can't switch to it for a couple OTP releases (because it doesn't exist in OTP20). You could use a macro to wallpaper over that and silence the warnings?

tomciopp:
What do you think about doing a version check within the code?

`erlang:display(erlang:system_info(otp_release))` should return the major version number. We can use that to use handshake if the version is greater than or equal to 21.

arjan:
The usual way to do it if to use a `platform_define` in rebar, that way you do the check compile-time and not in runtime.

Could you split this PR in 2? The `rand:uniform` change could be one, the conditional `ssl:accept` the other.

tomciopp:
@arjan @Vagabond Split PR into #172 and #173

PR #157, opened 2018-08-21
Use ranch as socket acceptor

seriyps:
## Motivation

Ranch is batle-tested high-performance socket acceptor pool. It's widely adopted and have lots of options to tune performance and to control overload.
gen_smtp development might be focused more on smtp implementation, not low-level networking.

## Downsides

It requires at least ranch 1.6, because of `ranch_ssl:handshake/3` API. That might be a problem for cowboy 1.x users.

It breaks some backward-compatibility:

* `gen_smtp_server:start_link/*` functions are removed
* `stop/1`, `sessions/1` 1st argument is not Pid, but listener name
* `gen_smtp_server_session` behaviour's callback `init/4` used to have number of connections as 2nd argument. This is not supported anymore. It was used mainly to implement overload protection, but itnow can be done by ranch `max_connections`

## What was done

gen_smtp_server socket acceptor was replaced by ranch.

gen_smtp's `socket` module is not used for server anymore.

It adds new API as a replacement for `gen_smtp_server:start_link/*`: `gen_smtp_server:child_spec/3` to use gen_smtp in embedded mode.

New `start/3` option `ranch_opts` was added. 

It also adds typespecs for server options and defines some missing `gen_smtp_server_session` `-callback` definitions (`init/4` and optional `handle_AUTH/4`).

It's kind of similar to abandoned cowboy-protocol branch.

I can also add some smtp server docs to readme and docstrings. This PR might be considered as a 1st draft.


mworrell:
I like the idea of removing code and using a kind-of-standard acceptor library.

Going forward we should indeed settle on the ranch version used in Cowboy 2. 

Having said that. There are quite a bit of (Elixir and Erlang) projects using Cowboy 1.
Is it possible to have a (rebar) switch to either use the older ranch or our own acceptor pool?

seriyps:
I think I may try to make it work with older Ranch versions (see https://github.com/ninenines/ranch/issues/139#issuecomment-187675843). Maybe even with run-time feature-detection (like `erlang:function_exported(ranch_ssh, handshake, 3)`).

I'm not sure what exactly will happen if some project will have both dependencies from cowboy 1.x and gen_smtp with ranch 1.6 dependency (I have my own project that uses gen_smtp and cowboy 1.x, but haven't tested it yet). However, it's always possible to overwrite the exact version of ranch in a top-level rebar.config with rebar3. And I don't know how mix and erlang.mk handles that.

So, I will try to make this PR work with both ranch 1.6 and ranch 1.3 (cowboy 1.x), if we agreed to proceed with ranch approach.

Vagabond:
I had a branch, long ago, that did this. I'm not sure why I never finished it.

Vagabond:
https://github.com/Vagabond/gen_smtp/tree/cowboy-protocol if anyone's interested.

seriyps:
Yep, I mentioned it in PR comment.

ср, 22 авг. 2018 г., 18:40 Andrew Thompson <notifications@github.com>:

seriyps:
Ok, now it's compatible with ranch 1.3.2 (cowboy 1.x) and Ranch 1.6. It made code a bit more messy, but still manageable.
To run tests with ranch 1.6 one can run `rebar3 eunit --module gen_smtp_server_session` and to run with ranch 1.3.2 run `rebar3 as ranch13 eunit --module gen_smtp_server_session `.

Also, server options were documented in README.

seriyps:
So, any feedback on this, @mworrell @Vagabond ?

Vagabond:
I'm in favor of this, although I have not reviewed the code yet.

mworrell:
Same here, less software is better, and ranch is well tested.

mworrell:
@seriyps Have you been running this in production?

mworrell:
@seriyps If I run `make test` on your branch then I get the following error:

```
TLS Started
EHLO from somehost.com
Mail from user@somehost.com
Mail to user@otherhost.com
message from user@somehost.com to [<<"user@otherhost.com">>] queued as b492ab6b8a2c40b5100c01472092231e, body length 85
............................................................................................................................F...........
Failures:

  1) socket:ssl_upgrade_test_/0: TCP connection can be upgraded to ssl
     Failure/Error: ?assertMatch({ sslsocket , _ , _ }, NewServerSocket)
       expected: = { sslsocket , _ , _ }
            got: {inet_async,#Port<0.81371>,675,{error,closed}}
     %% /Users/marc/Sites/misc/gen_smtp_seriyps/_build/test/lib/gen_smtp/src/socket.erl:637:in `socket:-ssl_upgrade_test_/0-fun-4-/0`
```

Is this expected? Maybe an issue with an old(er) test?

mworrell:
In testing I see that we have the same error on the current master.

@Vagabond shall we merge this and then look into the reported error separately?

seriyps:
@mworrell no, I haven't tried it in production. Maybe I should. I made this PR to check how the idea will be accepted by community in the first place. But now when I see that it have a chance to be accepted, I can try to polish it a bit more! Let's hold on for a little while, I will try to run it in prod

mworrell:
Thanks!  I definitely am interested to have this accepted, as it removes code and replaces it with well tested code that is in use by many other projects.

seriyps:
I created a new PR #171. It's almost the same as this one, but rebased over the latest changes in master

arjan:
See #171

PR #170, opened 2019-05-27
Fix ssl upgrade test failures

seriyps:
There were 2 tests faling because of race-condition between user-generated message and inet driver message:

```
  1) smtp_socket:ssl_upgrade_test_/0: TCP connection can be upgraded to ssl                                                                                                                                                     
     Failure/Error: ?assertMatch({ sslsocket , _ , _ }, NewServerSocket)                                         
       expected: = { sslsocket , _ , _ }                                                                                                                                                                                        
            got: {inet_async,#Port<0.397>,1,{error,closed}}                                                                                                                                                                     
     %% eunit_proc.erl:325:in `eunit_proc:with_timeout/3`                                                                
     Output:  
```

```
  2) smtp_socket:ssl_upgrade_test_/0: SSL client connection can't be upgraded again                                                                                                                                             
     Failure/Error: {error,function_clause,                                                                                                                                                                                     
                        [{ssl,close,                                                                                                                                                                                            
                             [{inet_async,#Port<0.401>,1,{error,closed}}],                                                                                                                                                      
                             [{file,"ssl.erl"},{line,294}]},                                                                                                                                                                    
                         {eunit_test,run_testfun,1,                                                                                                                                                                             
                             [{file,"eunit_test.erl"},{line,71}]},                                               
                         {eunit_proc,run_test,1,                                                                                                                                                                                
                             [{file,"eunit_proc.erl"},{line,510}]},                                                                                                                                                             
                         {eunit_proc,with_timeout,3,                                                                     
                             [{file,"eunit_proc.erl"},{line,335}]},                                                                                                                                                             
                         {eunit_proc,handle_test,2,                                                                                                                                                                             
                             [{file,"eunit_proc.erl"},{line,493}]},                                                                                                                                                              
                         {eunit_proc,tests_inorder,3,                                                                                                                                                                           
                             [{file,"eunit_proc.erl"},{line,435}]},                                                                                                                                                             
                         {eunit_proc,with_timeout,3,                                                                                                                                                                             
                             [{file,"eunit_proc.erl"},{line,325}]},                                                                                                                                                             
                         {eunit_proc,run_group,2,                                                                                                                                                                               
                             [{file,"eunit_proc.erl"},{line,549}]}]} 
```

Now fixed by using selective receive.

Is there any chance Travis builds will be enabled for gen_smtp?

PR #171, opened 2019-05-27
Use ranch as socket acceptor v2

seriyps:
I decided to create a new PR because I did quite big rebase over the latest changes (especially with "socket" module). Previous PR is #157. I did no code changes compared to previous one, just rebase.

All tests and dialyzer passes.

It also includes changes from #170, so, if #170 will be merged first, this PR will become 2 commits smaller.

Please, let me know if I should close this PR and update (force-push) previous one instead.


src/gen_smtp_server_session.erl
```diff
@@ -68,119 +74,141 @@
...
--spec start(Socket :: port(), Module :: atom(), Options :: [tuple()]) -> {'ok', pid()} | 'ignore' | {'error', any()}.
-start(Socket, Module, Options) ->
-	gen_server:start(?MODULE, [Socket, Module, Options], []).
+ranch_init({Ref, Socket, Transport, {Callback, RanchVer, Opts}}) ->
+	{ok, Socket} = ranch_handshake(RanchVer, Ref, Socket),
+	case init([Ref, Transport, Socket, Callback, RanchVer, Opts]) of
+		{ok, State, Timeout} ->
+			gen_server:enter_loop(?MODULE, [], State, Timeout);
+		{stop, Reason} ->
+			exit(Reason)
```
seriyps: FIXME: handle `ignore`

src/gen_smtp_server_session.erl
```diff
@@ -874,6 +923,32 @@ check_bare_crlf(Binary, _Prev, Op, Offset) ->
...
 	end.
 
+ranch_handshake(gte16, Ref, _Sock) ->
+	ranch:handshake(Ref);
+ranch_handshake(lt16, Ref, Sock) ->
+	ok = ranch:accept_ack(Ref),
+	{ok, Sock}.
+
+send(#state{transport = Transport, socket = Sock}, Data) ->
+    ok = Transport:send(Sock, Data).
```
seriyps: I'm often getting `@gen_smtp_server_session:send:933 gen_server <0.9016.2> terminated with reason: no match of right hand value {error,closed} in gen_smtp_server_session:send/2 line 933` in production. I believe it's because SMTP server clients are closing the socket without waiting for response from us. It's kind of ok, but maybe we should handle such cases explicitly? Because now it spams the logs.

Before this PR we were just ignoring such errors by not matching on smtp_socket:send return values

mworrell: If we received all info and did report the success to the remote client, then it is ok that remote drops the connection without properly saying goodbye.  So our behavior here really depends on the state we are in, if we got success with receiving the message then we should ignore the send-error, if we are in the middle of the send hand shake then we could log something and also give up. In the end it is up to the remote to try again.

seriyps: @mworrell another option is to define some `terminate` optional behavior callback

mworrell: Good idea, then we can log in the `terminate` if we are terminating in the middle of some transaction or just be silent if all was received and well.

The send code could just stop the `gen_server` if an error occurs.

seriyps: I tried to fix this, but it seems to require some major rewrite (eg, get rid of separate process to receive message body). I think I will revert this `ok = ` pattern for now (so it will behave the same as before my PR) and will make a separate PR to clean this up later.
What do you think?

src/gen_smtp_server_session.erl
seriyps: fixed

src/gen_smtp_server_session.erl
seriyps: So yep, I disabled send result checks for now and will add it in another PR

seriyps:
2 weeks in production for both ipv4 and ipv6; no issues found

seriyps:
Any news on this?

juhlig:
I was just about suggesting Ranch when I saw your PR here.

Please note that the upcoming Ranch 2.0, which should be available as a pre-release soon (and can already be used via the master branch), removes the functions that are deprecated now, so your implementation is not compatible with it.

However, making it compatible with both Ranch 1.x and 2.0 is easy. There is already a PR pending for `cowboy` [here](https://github.com/ninenines/cowboy/pull/1382) that enables it to work with both. Only four files needed to be changed, and only the changes in two of them (`cowboy_clear` and `cowboy_tls`) are relevant to `gen_smtp`. In a nutshell, you should create and export both `start_link/4` (for Ranch 1.x) and `start_link/3` (for Ranch 2.0), with equivalent behavior. Depending on what version of Ranch is used, either the one or the other will be called.

juhlig:
As far as that goes, even the `ranch_handshake/3` function is not necessary. `ranch:accept_ack/1` could be used in `start_link/4` (it is deprecated in favor of `ranch:handshake/1`, but won't be removed in any future 1.x version), and `start_link/3` would use `ranch:handshake/1` instead.

seriyps:
thanks! I will try to implement fixes that you suggested and test via ranch master branch.

seriyps:
@juhlig Ranch 2.0 support was added. Tested with ranch master

mworrell:
@arjan ok with a merge of this?  I think it is a good change to settle on ranch as that is used by many projects and makes the gen_smtp code smaller.

This might also be a reason to call the next to 1.x.  Especially as this code is stable and in production, so no reason to keep this at 0.x

seriyps:
It actually might make sense to tag one release before this is merged just to have a release with all the latest changes/fixes but without ranch for those who don't want it for any reason

mworrell:
I will make a milestone for the next 0.x and one for 1.0.  Then we can assign issues and MRs to those two milestones.

juhlig:
@essen is probably interested to hear this :)

mworrell:
@seriyps the ssl_handshake has been merged, so I guess that forces a rebase on your work :-(

seriyps:
@mworrell I rebased over the current master

mworrell:
Great, then we are standing ready for the 1.0 merges. That is, after the 0.15 has been released.
I added a 0.15 issue #185 for that release, and assigned it to @arjan as I am not familiar with the release procedure.

seriyps:
Wow! :champagne:

mworrell:
As we said, it is 1.0 merge time! 🎉

essen:
Cheers!

mworrell:
@essen we are becoming essen-software compatible 👍

PR #172, opened 2019-05-29
chore(Compiler Warning): Change from crypto:rand_uniform/2 to rand:un…

tomciopp:
…iform/1.

Splitting code changes from #156 into two PRs based on suggestion by @arjan.


seriyps:
Module `rand` was added in OTP 18, but in another PR you have `platform_define` even for R16.
But I think it should be ok to support only OTP >= 18, just `.travis.yml` should be updated (however, travis is not enabled now)

arjan:
Merging since we now require at least OTP 18

PR #173, opened 2019-05-29
chore(Compiler Warnings): Use ssl:handshake instead of ssl:accept for…

tomciopp:
… OTP versions >= 21

@arjan I'm not 100% if this syntax is correct or if it can be simplified, I'm relying on syntax highlighting from my editor and other examples from the repo. I will be happy to fix if anything looks wrong.

### DIFF
--- a/src/smtp_socket.erl
+++ b/src/smtp_socket.erl
@@ -120,13 +120,29 @@ accept(Socket, Timeout) when is_port(Socket) ->
 		Error ->
 			Error
 	end;
+
+
 accept(Socket, Timeout) ->
 	case ssl:transport_accept(Socket, Timeout) of
 		{ok, NewSocket} ->
-			ssl:ssl_accept(NewSocket),
+			ssl_handshake(NewSocket),
 			{ok, NewSocket};
 		Error -> Error
 	end.
+
+-ifdef(deprecated_ssl_accept).
+ssl_handshake(Socket) ->
+  ssl:ssl_accept(Socket);
+
+ssl_handshake(Socket, Options, Timeout) ->
+  ssl:ssl_accept(Socket, Options, Timeout).
+-else.
+ssl_handshake(Socket) ->
+  ssl:handshake(Socket);
+
+ssl_handshake(Socket, Options, Timeout) ->
+  ssl:handshake(Socket, Options, Timeout).
+-endif.
 
 -spec send(Socket :: socket(), Data :: binary() | string() | iolist()) -> 'ok' | {'error', any()}.
 send(Socket, Data) when is_port(Socket) ->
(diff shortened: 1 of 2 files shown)


src/smtp_socket.erl
```diff
@@ -120,13 +120,24 @@ accept(Socket, Timeout) when is_port(Socket) ->
...
 	end.
+-else.
+accept(Socket, Timeout) ->
+	case ssl:transport_accept(Socket, Timeout) of
+		{ok, NewSocket} ->
+			ssl:handshake(NewSocket),
+			{ok, NewSocket};
+		Error -> Error
+	end.
+-endif.
```
seriyps: another option would be to create a new function and use it in both places:

```
-ifdef(deprecated_ssl_accept).
ssl_handshake(Sock) ->
    ssl:handshake(Sock).
-else.
ssl_handshake(Sock) ->
    ssl:ssl_accept(NewSocket).
-endif.
```
This way you will have less code duplication.

arjan: Yes I agree with @seriyps here, less code duplication = better

mworrell:
@tomciopp We are preparing a 0.15 release, are you able to make the proposed changes and rebase?

mworrell:
I think this will give a merge conflict with #171, so we can do two things:

 1. Add this to a separate 0.x branch (which kind of makes sense)
 2. Add this to the master branch and then rebase #171 

@seriyps I think option (1) makes most sense, what do you think?

seriyps:
Well, conflict is not a problem for me. I can rebase / fix it, it's ok. So, up to you.

tomciopp:
@mworrell I can take care of this later today, or early tomorrow morning.

tomciopp:
@mworrell @seriyps @arjan Updated code so that there is a single function called for both situations. Let me know if anything else needs to change.

mworrell:
Looks good to me as well. Merging.

Thanks @tomciopp!

src/smtp_socket.erl
```diff
@@ -120,14 +120,30 @@ accept(Socket, Timeout) when is_port(Socket) ->
 		Error ->
 			Error
 	end;
+
+
 accept(Socket, Timeout) ->
 	case ssl:transport_accept(Socket, Timeout) of
 		{ok, NewSocket} ->
-			ssl:ssl_accept(NewSocket),
+			ssl_handshake(NewSocket),
```
lazedo: in OTP 21, handshake always returns a new Socket and we are dropping it here ?

mworrell: Thanks, see https://github.com/gen-smtp/gen_smtp/pull/187 if that would solve this.  (All tests pass, so apparently at the moment the ssl routines return the same socket)

PR #184, opened 2019-07-19
Add option for unlimited message size.

### LINKED ISSUE #4: Set max message size to 0 to allow unbounded messages
You currently can't allow messages of unlimited size like, eg. Postfix does, when you set the max to 0.

mworrell:
Fixes #4 

Also clean up some code and catch close socket errors during send tests.

To do:

 - [x] Add tests for testing message size limit

### DIFF
--- a/src/gen_smtp_server_session.erl
+++ b/src/gen_smtp_server_session.erl
@@ -89,7 +89,7 @@
 	{ok, Banner :: iodata(), CallbackState :: state()} | {stop, Reason :: any(), Message :: iodata()} | ignore.
 -callback code_change(OldVsn :: any(), State :: state(), Extra :: any()) ->  {ok, state()}.
 -callback handle_HELO(Hostname :: binary(), State :: state()) ->
-    {ok, pos_integer(), state()} | {ok, state()} | error_message().
+    {ok, pos_integer() | 'infinity', state()} | {ok, state()} | error_message().
 -callback handle_EHLO(Hostname :: binary(), Extensions :: list(), State :: state()) ->
     {ok, list(), state()} | error_message().
 -callback handle_AUTH(AuthType :: login | plain | 'cram-md5', Username :: binary(),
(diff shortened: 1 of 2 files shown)


mworrell:
@arjan @seriyps could you have a look at this?

src/gen_smtp_server_session.erl
```diff
@@ -66,6 +66,7 @@
 		ranch_version :: lt16 | gte16,
 		envelope = undefined :: 'undefined' | #envelope{},
 		extensions = [] :: [{string(), string()}],
+		maxsize = ?MAXIMUMSIZE,
```
seriyps: would be nice to have a typespec. `:: non_neg_integer()` should be good.

src/gen_smtp_server_session.erl
```diff
@@ -342,10 +326,14 @@ handle_request({<<"HELO">>, <<>>}, State) ->
 handle_request({<<"HELO">>, Hostname},
 			   #state{options = Options, module = Module, callbackstate = OldCallbackState} = State) ->
 	case Module:handle_HELO(Hostname, OldCallbackState) of
+		{ok, 0, CallbackState} ->
```
seriyps: just opinion: might it be better to use some atom as switch for "unlimited"? We can use `infinity` for example (similar to Erlang's built-in `timeout()` datatype).

seriyps: Also, this typespec needs to be updated now:

https://github.com/gen-smtp/gen_smtp/blob/628863f0bfd4439678bf216da5c1f55626ec2a8e/src/gen_smtp_server_session.erl#L90-L91

because it says `pos_integer()`

mworrell: I like the idea of using `infinity` - that also prevents accidental `0` configs.

src/gen_smtp_server_session.erl
mworrell: Yes, I changed it to `pos_integer() | 'infinity'`

Note that the callback server code has to return `{"SIZE", "0"}` and know to map `infinity` to `0`.  I guess that is ok, as you have to be prepared to handle unlimited size anyway.

seriyps:
Would be cool to finally merge this and push a new release!

PR #187, opened 2019-07-19
Make ssl_handshake similar across OTP versions

mworrell:
As mentioned by @lazedo in https://github.com/gen-smtp/gen_smtp/pull/173/files#r305487660 we are dropping a return value when using ssl:ssl_handshake functions.

This fixes that problem by making the newly introduced ssl_handshake function return same types across OTP versions.


mworrell:
Thanks @lazedo for pointing this out.

PR #209, opened 2020-05-16
smtp_server: allow any TLS options when STARTTLS

seriyps:
This is to allow any TLS options to be specified for server's STARTTLS. My usecase is quite specific: to serve multiple TLS hosts on single server, so, server can provide the sertificate that matches the host provided by client. But any other option can be specified as well, such as TLS versions or cipher suites.

Patch is bigger than one might expect mainly because I added a script to re-generate test TLS certificates and regenerated them, changing the names. So, all the paths had to be updated.


seriyps:
Green build: https://travis-ci.org/github/seriyps/gen_smtp/builds/688084478

mworrell:
I see this is backwards compatible with the old key/certfile options.
Great to be able to add more options.

PR #210, opened 2020-05-17
Use hut instead of io:format/2

seriyps:
I just replaced all `io:format/2` calls with hut's `?log(debug, ...)`.

Hut is a macro-only library with configurable logging backends https://github.com/tolbrino/hut


seriyps:
https://travis-ci.org/github/seriyps/gen_smtp/builds/688082128 travis is green

mworrell:
Much better indeed.

Solves the problem of the `logger` vs SASL vs `lager` situation.

PR #212, opened 2020-05-18
Make it compatible with OTP-23; add OTP-23 to travis. Fixes #207

### LINKED ISSUE #207: Deprecated crypto:hmac/3 in otp 23
```
smtp_util.erl:73: Warning: crypto:hmac/3 is deprecated and will be removed in OTP 24; use use crypto:mac/4 instead
```
```erlang
Bin = crypto:mac(hmac, md5, Key, Data),
```

seriyps:
This one fixes #207 .Another alternative would be to add `platform_define` to rebar.config but this ?OTP_RELEASE seems more universal. I need 2 `-ifdef`'s because ?OTP_RELEASE was introduced in OTP-21 and we support versions from OTP-18


seriyps:
Green build https://travis-ci.org/github/gen-smtp/gen_smtp/builds/688498007
BTW, any ideas why travis is not linked automatically?

mworrell:
Great!

PR #221, opened 2020-09-30
Fix/rebar dependency loosen

arjan:
Compiling `swoosh` upstream yields an error:

```
Failed to use "ranch" because
  gen_smtp (versions 1.0.0 and 1.0.1) requires 1.6.2
  mix.lock specifies 1.7.1

** (Mix) Hex dependency resolution failed, change the version requirements of your dependencies or unlock them (by using mix deps.update or mix deps.unlock). If you are unable to resolve the conflicts you can try overriding with {:dependency, "~> 1.0", override: true}
```

This branch uses looser version specifiers in `rebar.config` for our dependencies.

### DIFF
--- a/rebar.config
+++ b/rebar.config
@@ -23,8 +23,8 @@
   {smtp_rfc822_parse, return_error, 2}
  ]}.
 
-{deps, [{ranch, "1.6.2"},
-        {hut, "1.3.0"}]}.
+{deps, [{ranch, "~> 1.6"},
+        {hut, "~> 1.3"}]}.
 
 {profiles,
  [
@@ -32,7 +32,7 @@
    [
     {deps,
      [
-      {eiconv, "1.0.0"}
+      {eiconv, "~> 1.0"}
      ]
     },
     {dialyzer,
@@ -51,16 +51,16 @@
      ]}
    ]},
   {ranch13,
-   [{deps, [{ranch, "1.3.2"}]}]},
+   [{deps, [{ranch, "~> 1.3"}]}]},
   {ranch20,
-   [{deps, [{ranch, "2.0.0"}]}]},
-  {test, 
+   [{deps, [{ranch, "~> 2.0"}]}]},
+  {test,
    [
     {cover_enabled, true},
     {cover_print_enabled, true},
     {deps,
      [
-      {eiconv, "1.0.0"}
+      {eiconv, "~> 1.0"}
      ]}
    ]}
  ]}.


arjan:
@seriyps what do you think?

seriyps:
I guess it's fine. I didn't know rebar deps supports this syntax!

ср, 30 сент. 2020 г., 17:26 Arjan Scherpenisse <notifications@github.com>:

seriyps:
I was not able to use ranch 1.7.1 as dependency because it does not support
OTP18 and we have 18 in our Travis. I guess now when 1.0 is released we can
drop OTP18 and upgrade ranch to 1.7

ср, 30 сент. 2020 г., 19:11 Сергей Прохоров <seriy.pr@gmail.com>:

arjan:
Ah yes indeed. The current rebar.lock still has the 1.6 dependency but the rebar.config has the looser specification now, which (hopefully) will solve upstream compilation. So we can keep it like this and still support OTP18, you agree?

seriyps:
It's an interesting situation =)

gen_smtp supports Ranch versions from 1.3 to 2.0 with special handling for versions:

- `< 1.6`: (in order to support Cowboy 1.x), because ranch API have changed
- `1.6 <= RanchVer < 1.7`: because ranch 1.7+ does not support OTP-18, but we do
- `>= 2.0`: because ranch 2.0 have some API changes

I really believe that now, when we did 1.0 release we can drop OTP-18 support, because those who need both OTP-18 and ranch could use 1.0 tag; those who need Cowboy 1.x and ranch (on any OTP version) could also hook to 1.0.

Problems that OTP-18 making for us now is that we can't test OTP-18 with ranch >= 1.7 and we also have some compile-time hacks to support it (see, eg, https://github.com/gen-smtp/gen_smtp/search?q=OTP_18)

seriyps:
We could technically set `{deps, [{ranch, ">= 1.3.2"}]}` (or what's the syntax to say "any version above or equal to 1.3.2, even 2.x is fine"), if we drop OTP-18.

seriyps:
> or what's the syntax to say "any version above or equal to 1.3.2, even 2.x is fine"

Seems `^1.3.2 || ^2` should do the trick: https://jubianchi.github.io/semver-check/#/constraint/^1.3.2%20||%20^2

gmile:
Would love to see this PR accepted. Hitting the same issue with dependency version mismatch.

seriyps:
@arjan I think we can merge it as is? And I'm going to make a PR to get rid of OTP-18 support (it's ~4 years old!)

seriyps:
I believe this should be fixed by #225 (it introduced `^1.6.2 || ^2`)

PR #223, opened 2020-10-26
Fix test compilation warnings + 1 test

### LINKED ISSUE #222: eunit error in spawn'ed process
I'm getting
```erlang
{{badmatch,{error,closed}},
 [{smtp_socket,'-ssl_upgrade_test_/0-fun-6-',1,
               [{file,"STUFF/gen_smtp/src/smtp_socket.erl"},
                {line,707}]}]}
```
This happens inside a spawn'ed process which doesn't break the `eunit` "suite".
Is this intended behaviour?

paulo-ferraz-oliveira:
[Fix #222]

### DIFF
--- a/rebar.config
+++ b/rebar.config
@@ -56,6 +56,7 @@
    [{deps, [{ranch, "2.0.0"}]}]},
   {test, 
    [
+    {erl_opts, [nowarn_export_all]},
     {cover_enabled, true},
     {cover_print_enabled, true},
     {deps,
--- a/rebar.config
+++ b/rebar.config
@@ -56,7 +56,6 @@
    [{deps, [{ranch, "2.0.0"}]}]},
   {test, 
    [
-    {erl_opts, [nowarn_export_all]},
     {cover_enabled, true},
     {cover_print_enabled, true},
     {deps,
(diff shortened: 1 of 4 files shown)


rebar.config
```diff
@@ -56,6 +56,7 @@
    [{deps, [{ranch, "2.0.0"}]}]},
   {test, 
    [
+    {erl_opts, [nowarn_export_all]},
```
seriyps, line 59: Could you, please, just replace `-compile([export_all]).` with `-compile([export_all, nowarn_export_all]).` in test module's code instead of disabling this warning for all the modules?

seriyps (changes requested):
Thanks! Nice catch!

rebar.config
paulo-ferraz-oliveira: Sure. It's only for "all modules in profile `test`" but I can do that.

paulo-ferraz-oliveira: 06be1b8

seriyps (approved):
Thanks! I'll wait one day if anyone else has any comments and then merge

arjan:
This looks great, thanks @paulo-ferraz-oliveira!

PR #224, opened 2020-10-28
Drop support for OTP-18

seriyps:
OTP-18 apparently is not that much of effort to still support, but I think it just doesn't make sense. OTP-18.3 was released 14 Mar 2016 and is not supported by OTP team anymore.

If someone still needs it, hey can stick to gen_smtp 1.0


arjan:
:+1:

mworrell (approved):

PR #225, opened 2020-10-28
Drop support for Ranch below 1.6.0

seriyps:
Again, those who need older Ranch (eg, they use Cowboy 1.x) can use gen_smtp 1.0

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -20,8 +20,7 @@ Also included is a MIME encoder/decoder, sorta according to RFC204{5,6,7}.
 
 IPv6 is also supported (at least serverside).
 
-SMTP server uses ranch as socket acceptor. It can use Ranch 1.3 to be compatible
-with cowboy 1.x, but Ranch 1.6+ is preferred.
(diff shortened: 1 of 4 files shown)


arjan:
Looks good. Next is dropping Erlang < 19? :-)

mworrell:
We are now at Erlang 23, so I am good with dropping < 21.
Then we support the last three releases.

seriyps:
I already created a PR to drop OTP-18. But OTP-19 is 0-effort to support right now, so, dropping it will be just removing 1 line from .travis.yml.
Dropping OTP <21 will help a bit (we have some `-ifdef`'s for pre-21 versions; also we are only able to build edocs on OTP-21+), but effort to support OTP-19+ is not that high so far.

aptinio:
I tried adding [master](/gen-smtp/gen_smtp/tree/master) as a dependency in an Elixir project and I got:

```
(Hex.Version.InvalidRequirementError) invalid requirement: "^1.6.2 || ^2"
```

I'm not familiar with rebar but I wonder if using `~>` instead of `^` would help.

arjan:
Thanks @aptinio, trying to fix this in #227.

PR #226, opened 2020-10-30
Smtp server - better error handling

seriyps:
4 independent commits, might be easier to review one by one.

This way we let callback module to decide how to handle such errors. All 3 are actually non-recoverable, as far as I understand, but we didn't crash on `send` historically.

We will stop the session process anyway (callback module can avoid crash logs by returning `{stop, normal, State}` from the `handle_error` function). For `send` error we are changing previous behaviour (used to ignore, now are going to stop).

### DIFF
--- a/src/gen_smtp_server_session.erl
+++ b/src/gen_smtp_server_session.erl
@@ -976,8 +976,13 @@ send(#state{transport = Transport, socket = Sock}, Data) ->
     %% TODO: handle send errors
     Transport:send(Sock, Data).
 
-setopts(#state{transport = Transport, socket = Sock}, Opts) ->
-    ok = Transport:setopts(Sock, Opts).
+setopts(#state{transport = Transport, socket = Sock} = St, Opts) ->
+    case Transport:setopts(Sock, Opts) of
+		ok -> ok;
+		{error, Err} ->
+			St1 = handle_error(setopts, Err, St),
+			throw({stop, {setopts_error, Err}, St1})
+	end.
 
 hostname(Opts) ->
     proplists:get_value(hostname, Opts, smtp_util:guess_FQDN()).
--- a/src/gen_smtp_server_session.erl
+++ b/src/gen_smtp_server_session.erl
@@ -831,7 +831,7 @@ try_auth(AuthType, Username, Credential, #state{module = Module, envelope = Enve
 					{ok, NewState}
 				end;
 		false ->
-			error_logger:error_msg("Please define handle_AUTH/4 in your server module or remove AUTH from your module extensions~n"),
+			?log(warning, "Please define handle_AUTH/4 in your server module or remove AUTH from your module extensions~n"),
 			send(State, "535 authentication failed (#5.7.1)\r\n"),
 			{ok, NewState}
 	end.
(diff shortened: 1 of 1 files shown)


src/gen_smtp_server_session.erl
```diff
@@ -972,12 +979,21 @@ check_bare_crlf(Binary, _Prev, Op, Offset) ->
...
 
-setopts(#state{transport = Transport, socket = Sock}, Opts) ->
-    ok = Transport:setopts(Sock, Opts).
+setopts(#state{transport = Transport, socket = Sock} = St, Opts) ->
+    case Transport:setopts(Sock, Opts) of
+		ok -> ok;
+		{error, Err} ->
+			St1 = handle_error(setopts_error, Err, St),
+			throw({stop, {setopts_error, Err}, St1})
+	end.
```
seriyps, line 996: it used to fail with smth like
```
[2020-10-29T05:16:17.107810+01:00 error <0.27720.227> proc_lib:crash_report/4:508]
crasher:
 initial call: gen_smtp_server_session:ranch_init/1,
 pid: <0.27720.227>,
 registered_name: [],
error: {{badmatch,{error,closed}},[{gen_smtp_server_session,setopts,2,[{file,"<..>/_build/default/lib/gen_smtp/src/gen_smtp_server_session.erl"},{line,993}]},
{gen_smtp_server_session,handle_info,2,[{file,"<..>/_build/default/lib/gen_smtp/src/gen_smtp_server_session.erl"},{line,283}]},
{gen_server,try_dispatch,4,[{file,"gen_server.erl"},{line,637}]},
{gen_server,handle_msg,6,[{file,"gen_server.erl"},{line,711}]},
{proc_lib,init_p_do_apply,3,[{file,"proc_lib.erl"},{line,249}]}]},
ancestors: [<0.1007.0>,<0.1006.0>,ranch_sup,<0.919.0>],
message_queue_len: 1,
messages: [{ssl_closed,{sslsocket,{gen_tcp,#Port<0.6554831>,tls_connection,undefined},[<0.21175.227>,<0.18409.227>]}}],
links: [<0.1007.0>],
 dictionary: [],
 trap_exit: false,
 status: running,
 heap_size: 121536,
 stack_size: 27,
 reductions: 300084;
 neighbours:
```
Now `handle_error` callback will be called first and it can return `{stop, normal, ...}` to avoid SASL crash logs

src/gen_smtp_server_session.erl
```diff
@@ -908,8 +915,8 @@ receive_data(Acc, Transport, Socket, RecvSize, Size, MaxSize, Session, Options)
 		{error, timeout} ->
 			receive_data(Acc, Transport, Socket, 0, Size, MaxSize, Session, Options);
 		{error, Reason} ->
-			error_logger:error_msg("SMTP receive error: ~p~n", [Reason]),
-			exit(receive_error)
```
seriyps, line 912: those were propagated (via link) to a session process and the session process is "abruptly" killed:
```
[2020-10-29T08:00:52.750562+01:00 error <0.21202.231>] SMTP receive error: closed
[2020-10-29T08:00:52.751746+01:00 error <0.1007.0>] Ranch listener {gen_smtp_server,{0,0,0,0}} had connection process started with gen_smtp_server_session:start_link/4 at <0.19039.231> exit with reason: receive_error
```

Now this message is handled around line 230, callback module could decide what to do in the same way as for `setopts`. Session process will stop anyway by returning `{stop, ...}` now instead of just dying, which is good, because, eg, `terminate` gen_server callback will be called as well.

seriyps:
Hi. May I please get some review on this?

mworrell (approved):
Looks perfect to me, also fixes the noisy errors in our logs.

seriyps:
Should I merge it then? Or you'll do? Or need to wait for @arjan? Seems we haven't agreed on the procedure…

arjan:
I think a :+1: from either of us should do it :-) I have not been able to look at this in detail, sorry.

mworrell:
Yes, if we had two people checking it, then it should be ok to merge.

Merging 👍

PR #227, opened 2020-11-01
Fix ranch deps specification so it works with Elixir

arjan:
See #225; given specification did not work with Elixir somehow

### DIFF
--- a/rebar.config
+++ b/rebar.config
@@ -22,8 +22,7 @@
   {smtp_rfc822_parse, return_error, 2}
  ]}.
 
-{deps, [{ranch, "^1.6.2 || ^2"},
-        {hut, "1.3.0"}]}.
+{deps, [{ranch, ">= 1.6.2 or ~> 2.0"}, {hut, "~> 1.3"}]}.
 
 {profiles,
  [
@@ -51,7 +50,7 @@
    ]},
   {ranch20,
    [{deps, [{ranch, "2.0.0"}]}]},
-  {test, 
+  {test,
    [
     {cover_enabled, true},
     {cover_print_enabled, true},
--- a/rebar.config
+++ b/rebar.config
@@ -22,7 +22,7 @@
   {smtp_rfc822_parse, return_error, 2}
  ]}.
 
-{deps, [{ranch, ">= 1.6.2 or ~> 2.0"}, {hut, "~> 1.3"}]}.
+{deps, [{ranch, "~> 1.6.2 or ~> 2.0"}, {hut, "~> 1.3"}]}.
 
 {profiles,
  [


rebar.config
```diff
@@ -22,8 +22,7 @@
   {smtp_rfc822_parse, return_error, 2}
  ]}.
 
-{deps, [{ranch, "^1.6.2 || ^2"},
-        {hut, "1.3.0"}]}.
+{deps, [{ranch, ">= 1.6.2 or ~> 2.0"}, {hut, "~> 1.3"}]}.
```
aptinio, line 25: `>= 1.6.2` would match 3.0 when it comes out.

Since we want to skip over 1.7, I think we need:
```
~> 1.6.2 or ~> 2.0
```
which is equivalent to:
```
(>= 1.6.2 and < 1.7.0) or (>= 2.0.0 < 3.0.0)

Comment (arjan) [reply rebar.config:25]: Good catch!

File: rebar.config [hunk 2]
```diff
@@ -22,8 +22,7 @@
   {smtp_rfc822_parse, return_error, 2}
  ]}.
 
-{deps, [{ranch, "^1.6.2 || ^2"},
-        {hut, "1.3.0"}]}.
+{deps, [{ranch, "~> 1.6.2 or ~> 2.0"}, {hut, "~> 1.3"}]}.
```
seriyps, line 25: Surprizingly, I can't find any specification for version range syntax. Found this cheatsheet https://devhints.io/semver and this online calculator: https://jubianchi.github.io/semver-check.
None of them know about `or`, but both mention `||`.

Any ideas how rebar3 and mix calculate version ranges? I checked rebar3 code but couldn't find anything semver releted

seriyps: Ok, seems rebar3 code is here https://github.com/erlang/rebar3/blob/5e2238707402fa6d149ca186494340ce7b9ee6a2/src/rebar_packages.erl#L162.

Maybe just use `>= 1.6.2`?

arjan: I only tested with mix and the `||` syntax did not work while `or` did...
what was wrong with ranch 1.7?

aptinio: Based on [this](https://github.com/gen-smtp/gen_smtp/pull/221#issuecomment-702076249), I think skipping over ranch 1.7 is no longer necessary now that OTP-18 support has been dropped.

pacoguzman:
Hi 👋 , I'm  interested in this, any progress?

wmnnd:
Without this patch, it’s currently impossible to install `gen_smtp` with a current version of Phoenix.
The version on Hex has too strict dependency requirements, the version on master is broken with Mix.

mworrell:
@arjan I am a bit confused between this merge request and #221. What is the difference and can we now use ranch > 1.7 (as suggested in one of the comments, as we don't support OTP18 anymore)

rebar.config
LatitudeAdjustment: It appears npm package manager uses "||" as logical-or whereas Hex package manager (mix) uses "or".

LatitudeAdjustment: It appears npm package manager uses "||" as logical-or whereas Hex package manager (mix) uses "or".

PR #228, opened 2020-11-17
Speed up quoted printable encoding

bossek:
Encoding to quoted printable was quite slow in cases when calling
string:substr and subsequent lists:concat for inserting soft newline was
used too often. New implementation uses only fast head operations.

### DIFF
--- a/src/mimemail.erl
+++ b/src/mimemail.erl
@@ -899,9 +899,7 @@ encode_quoted_printable(Body, Acc, L) when L >= 75 ->
 			encode_quoted_printable(Body, [$\n, $\r, $= | Acc], 0);
 		SIndex when (L - 75) < SIndex ->
 			% okay, we can safely stick some whitespace in
-			Prefix = string:substr(Acc, 1, SIndex-1),
-			Suffix = string:substr(Acc, SIndex),
-			NewAcc = lists:concat([Prefix, "\n\r=", Suffix]),
+			NewAcc = insert_soft_newline(Acc, SIndex - 1),
 			encode_quoted_printable(Body, NewAcc, 0);
 		_ ->
 			% worst case, we're over 75 characters on the line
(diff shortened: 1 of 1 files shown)


bossek:
For comparing before/after performance I used following:

```
Mail = fun(Body) ->
 {<<"multipart">>, <<"alternative">>,
  [{<<"From">>, <<"test@example.org">>}, {<<"To">>, <<"test@example.org">>}, {<<"Subject">>, <<"test">>}, {<<"Mime-Version">>, <<"1.0">>}], #{},
  [{"text", "html", [{<<"Content-Type">>, <<"text/html; charset=\"utf-8\"">>}, {<<"Content-Transfer-Encoding">>, <<"quoted-printable">>}],
  #{content_type_params => [{<<"charset">>, <<"utf-8">>}], disposition => <<"inline">>}, Body}]}
end.

timer:tc(fun() -> mimemail:encode(Mail(unicode:characters_to_binary(string:copies(" ř", 100000))), []) end).
timer:tc(fun() -> mimemail:encode(Mail(binary:list_to_bin(string:copies("a a", 100000))), []) end).
```

Both calls takes seconds before improvement and hundreds/tens of milliseconds after.

src/mimemail.erl
```diff
@@ -936,6 +932,11 @@ encode_quoted_printable(<<H, T/binary>>, Acc, L) ->
 	[A, B] = lists:flatten(io_lib:format("~2.16.0B", [H])),
 	encode_quoted_printable(T, [B, A, $= | Acc], L+3).
 
+insert_soft_newline([H | T], AfterPos) when AfterPos > 0 ->
+	[H | insert_soft_newline(T, AfterPos - 1)];
+insert_soft_newline(Str, 0) ->
+	[$\n, $\r, $= | Str].
```
seriyps, line 938: minor: I think you can do `"\n\r=" ++ Str` and compiler will be able to optimize it to what your code looks like now

seriyps (approved):
Nice, thanks!

mworrell:
Good work, thank you!

mworrell:
@seriyps Ready to be merged?

seriyps:
@mworrell yes

src/mimemail.erl
bossek: You are right! Should I change it before merge or keep it as it is?

seriyps: I'm fine as it is now.

bossek: OK, thanks!

mworrell:
@bossek Merging now, thank you for your contribution!

bossek:
@mworrell Glad to help :-)

PR #241, opened 2021-02-03
Require Ranch >= 1.7

arjan:
This should fix the issue with the too strict version requirement check.

cc @mworrell @seriyps


arjan:
We default now to ranch 2.0 in rebar.lock; apparently this won't work on OTP 19...

seriyps (approved):

arjan:
Unfortunately the tests now fail; maybe because it uses ranch 2.0 now? WDYT @seriyps ? I could put rebar.lock back to 1.7...

seriyps:
Yes, let's try to use 1.7 in rebar.lock

PR #243, opened 2021-02-18
Be explicit in our intentions (OTP 24 readiness)

paulo-ferraz-oliveira:
This pull request fixes two minor warnings that OTP 24 will present after https://github.com/erlang/otp/pull/2995.

### DIFF
--- a/src/gen_smtp_application.erl
+++ b/src/gen_smtp_application.erl
@@ -9,7 +9,7 @@ ensure_all_started(App) ->
 
 -spec start_ok(atom(), ok | {error, term()}) -> ok | {error, {term(), atom()}}.
 start_ok(_App, ok) -> ok;
-start_ok(_App, {error, {already_started, _App}}) -> ok;
+start_ok(App, {error, {already_started, App}}) -> ok;
 start_ok(App, {error, {not_started, Dep}}) ->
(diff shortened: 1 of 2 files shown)


mworrell:
Perfect, thanks!

src/gen_smtp_application.erl
```diff
@@ -9,7 +9,7 @@ ensure_all_started(App) ->
 
 -spec start_ok(atom(), ok | {error, term()}) -> ok | {error, {term(), atom()}}.
 start_ok(_App, ok) -> ok;
-start_ok(_App, {error, {already_started, _App}}) -> ok;
+start_ok(App, {error, {already_started, App}}) -> ok;
 start_ok(App, {error, {not_started, Dep}}) ->
```
seriyps, line 13: not related to this PR, but in general, I think we don't need this module anymore. `application:ensure_all_started` was added to OTP R16, and gen_smtp does not support such old releases anymore

https://github.com/gen-smtp/gen_smtp/commit/beeeef71fa655076fe17f07ce243e0ad2a83b486

paulo-ferraz-oliveira: You right. I think I did the same fix for `lager` one of these days.

PR #247, opened 2021-02-22
Mimemail proper

### LINKED ISSUE #246: mimemail: utf-8 headers with newline in them would cause a crash when encoding
```erlang
> Mail = {<<"text">>, <<"plain">>, [{<<"Subject">>, <<"Привет\nмир"/utf8>>}, {<<"From">>, <<"test@example.com">>}, {<<"Content-Type">>, <<"text/plain; charset=utf-8">>}], #{}, <<"Test">>}.
{<<"text">>,<<"plain">>,
 [{<<"Subject">>,
   <<208,159,209,128,208,184,208,178,208,181,209,130,10,208,
     188,208,184,209,128>>},
  {<<"From">>,<<"test@example.com">>},
  {<<"Content-Type">>,<<"text/plain; charset=utf-8">>}],
 #{},<<"Test">>}
> mimemail:encode(Mail).
** exception error: no function clause matching
 mimemail:rfc2047_utf8_encode([10,208,188,208,184,209,128],
                                                                             "5B=0D=2B=0D=8B=0D=08=1D=F9=0D=?Q?8-FTU?=",40,"28=1D=") (gen_smtp/src/mimemail.erl, line 964)
     in function  mimemail:encode_headers/1 (gen_smtp/src/mimemail.erl, line 788)
     in call from mimemail:encode/2 (gen_smtp/src/mimemail.erl, line 165)
```

here `10` is `\n`. I guess that's because we don't have clause for characters below 32 here, so, I guess, `\t` and `\r` would cause the same issue

https://github.com/gen-smtp/gen_smtp/blob/7b50cfa52620e296ee7e07e45ce550bfde6ef0f5/src/mimemail.erl#L964-L985

seriyps:
There are 2 commits: 1st one just fixes #246 (I have choosen, unlike Python, to do QP-encoding of `\r` and `\n`, because otherwise `mimemail` is not able to decode the result).

2nd commit is some "proper" property-based test suite for `mimemail`.

Fix for #246 and test suite were created together, but I decided t split to 2 commits, because they are technically unrelated.

It might be interesting to think about having some "proper" test suite where we compare encoding/decoding of `mimemail` with some other mime codec (eg, some PHP or Python one) to see if they can inter-communicate (mail encoded by Erlang can be read by Python and vice versa), but next time.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,6 +27,8 @@ jobs:
         run: make xref
       - name: Test
         run: make test
+      - name: Proper
+        run: make proper
       - name: Cover
         run: make cover
       - name: Dialyze
--- a/Makefile
+++ b/Makefile
@@ -10,6 +10,9 @@ test:
 	ERL_AFLAGS="-s ssl" 
 	./rebar3 eunit -c
 
+proper:
+	./rebar3 proper -c
+
 cover:
 	./rebar3 cover --verbose --min_coverage $(MINIMAL_COVERAGE)
 
--- a/rebar.config
+++ b/rebar.config
@@ -17,6 +17,10 @@
   deprecated_functions
  ]}.
 
+{project_plugins, [
+    rebar3_proper
+]}.
+
 {xref_ignores,
  [
   {smtp_rfc822_parse, return_error, 2}
@@ -57,7 +61,8 @@
     {cover_print_enabled, true},
     {deps,
      [
-      {eiconv, "1.0.0"}
+      {eiconv, "1.0.0"},
+      {proper, "1.3.0"}
      ]}
    ]}
  ]}.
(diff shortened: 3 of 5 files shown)


seriyps:
Is anyone available to review this? @arjan @mworrell ? `prop_mimemail` suite mainly just generates arbitrary `mimemail:mimetuple()` structures (plain as well as multipart) and

1) Just tries to encode them with `mimemail:encode`
2) Tries to encode and then cecode them with mimemail and checks that the result of encode+decode sequence is more or less the same as the original `mimetuple()`

seriyps:
Hey, may I do smth more to get the feedback / review?

mworrell (approved):
Good to add the proper testing as well!

seriyps:
Thank you @mworrell !

Issue #253, opened 2021-03-17
Bare LF


maennchen:
When sending emails to some SMTP servers, a result like this is replied:

```elixir
{:error, :send, {:temporary_failure, 'cbx.virusfree.cz', "451 See http://pobox.com/~djb/docs/smtplf.html.\r\n"}}
```

> You can't get mail through to msn.com and thousands of other systems around the Internet. Your mailer is violating 822bis section 2.3, which specifically prohibits all bare LFs.

I'll try to find out exactly what message is causing this behavior. This is however not that easy since I do not get the message logged together with the error at the moment.

seriyps:
do you generate email payload with `mimemail`, or it comes from somewhere else?

maennchen:
@seriyps I generate the email with `mimemail`.

```elixir
:mimemail.encode(
  {"text", "plain",
    [
      {"Subject", Changeset.fetch_field!(changeset, :subject)},
      {"From", :smtp_util.combine_rfc822_addresses([{from_name, from_email}])},
      {"To",
      :smtp_util.combine_rfc822_addresses([
        {to_name, Changeset.fetch_field!(changeset, :recipient)}
      ])},
      {"Auto-submitted", "yes"},
      {"X-Auto-Response-Suppress", "All"},
      {"Precedence", "auto_reply"},
      {"Date", Calendar.strftime(DateTime.utc_now(), "%a, %-d %b %Y %X %z")},
      {"Message-ID", Changeset.fetch_field!(changeset, :uuid)},
      {"Content-Type", "text/plain; charset=utf-8"},
      {"Content-Transfer-Encoding", "7bit"}
    ], [], Changeset.fetch_field!(changeset, :body)},
  encoding_options
)
```

seriyps:
Interesting. I did some extra testing in #247 and I see that `mimemail` indeed is able to generate results with "bare newlines" (`\n` without leading `\r`). Few examples:

```erlang
{<<"multipart">>,<<"mixed">>,
 [{<<"From">>,<<"test@example.com">>}],
 #{},
 [{<<"text">>,<<"html">>,[],
   #{content_type_params => [{<<"charset">>,<<"utf-8">>}],
     disposition => <<"inline">>},
   <<"<!doctype html><html><body><p>\n</p></body></html>">>}]}.

{<<"text">>,<<"plain">>,
 [{<<"From">>,<<"test@example.com">>}],
 #{},<<"\n">>}

[...]

PR #254, opened 2021-03-22
Verify and fix some of the RFC-2045 constraints of mimemail

seriyps:
Fixes:
* quoted-printable "very long lines" bug fixed
* never produce bare CR or LF (Fixes: #253)

Validations:
* no bare CR or LF
* no bytes above 127
* no 0-bytes
* no lines longer than ~1000 chars
* quoted-printable should not produce lines over 76 chars

### DIFF
--- a/src/mimemail.erl
+++ b/src/mimemail.erl
@@ -56,6 +56,7 @@
 -endif.
 
 -export([encode/1, encode/2, decode/2, decode/1, get_header_value/2, get_header_value/3, parse_headers/1]).
+-export([encode_quoted_printable/1, decode_quoted_printable/1]).
 
 -export_type([mimetuple/0,
               mime_type/0,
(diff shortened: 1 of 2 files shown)


src/mimemail.erl
```diff
@@ -652,91 +653,98 @@ check_headers([Header | Tail], Headers) ->
 
 ensure_content_headers(Type, SubType, Parameters, Headers, Body, Toplevel) ->
 	CheckHeaders = [<<"Content-Type">>, <<"Content-Disposition">>, <<"Content-Transfer-Encoding">>],
-	ensure_content_headers(CheckHeaders, Type, SubType, Parameters, lists:reverse(Headers), Body, Toplevel).
+	CheckHeadersValues = [{Name, get_header_value(Name, Headers)} || Name <- CheckHeaders],
+	ensure_content_headers(CheckHeadersValues, Type, SubType, Parameters, lists:reverse(Headers), Body, Toplevel).
 
 ensure_content_headers([], _, _, Parameters, Headers, _, _) ->
 	{Parameters, lists:reverse(Headers)};
-ensure_content_headers([Header | Tail], Type, SubType, Parameters, Headers, Body, Toplevel) ->
```
seriyps, line 659: no actual changes in this function. Just made it less nested.

src/mimemail.erl
```diff
@@ -746,14 +754,52 @@ guess_best_encoding(Body) ->
...
 	if
-		Percent == 100 ->
-			<<"7bit">>;
 		Percent > 80 ->
 			<<"quoted-printable">>;
 		true ->
 			<<"base64">>
 	end.
 
+%% https://tools.ietf.org/html/rfc2045#section-2.7:
```
seriyps, line 763: now there are much more strict requirements to allow body to be sent as is:

no lines longer than 998 bytes
no bare CR or LF
no 0 bytes or bytes above 127

src/mimemail.erl
```diff
@@ -900,7 +946,7 @@ encode_quoted_printable(Body, Acc, L) when L >= 75 ->
 		SIndex when (L - 75) < SIndex ->
 			% okay, we can safely stick some whitespace in
 			NewAcc = insert_soft_newline(Acc, SIndex - 1),
-			encode_quoted_printable(Body, NewAcc, 0);
+			encode_quoted_printable(Body, NewAcc, SIndex);
 		_ ->
```
seriyps, line 950: this and next diff are actually bugfixes. Here we used to set "current line length" to 0 even though the line we are building might be much longer than 0.

test/prop_mimemail.erl
```diff
@@ -99,6 +100,80 @@ match({TypeA, SubTypeA, HeadersA, _ParamsA, BodyA},
...
+	   Body,
+	   proper_types:oneof([?SIZED(Size, printable_ascii(Size * 50)),
+						   ?SIZED(Size, printable_ascii_and_cariage(Size * 50)),
+						   printable_ascii(),
+						   printable_ascii_and_cariage(),
+						   nonull_utf8(),
+						   proper_types:binary()]),
+	   begin
+		   [QPEncoded] = mimemail:encode_quoted_printable(Body),
+		   ?assertEqual(Trim(Body), Trim(mimemail:decode_quoted_printable(QPEncoded))),
```
seriyps, line 120: we still can loose heading and trailing whitespaces and tabs. Not sure if it's a bug, but doesn't sound right. Can be fixed in the following iterations.

mworrell:
Wow, that was quite a bit of work.... good additions, cleanup and fixes.

mworrell (approved):
Good work, thank you!

Issue #196, opened 2019-11-23
add LMTP support


Elemecca:
LMTP support in the server would be useful for people implementing any kind of mail-handling service that sits behind a standard MTA. LMTP ([RFC 2033](https://tools.ietf.org/html/rfc2033)) is a protocol derived from ESMTP used in local mail delivery applications (e.g. from an MTA like Postfix to a mailbox server like Dovecot). There are only a few differences from ESMTP, so it shouldn't be too difficult to implement within the existing framework.

First, to ensure everyone agrees which protocol they're speaking, LMTP disallows the `HELO` and `EHLO` commands and replaces them with `LHLO`, which has the same semantics as `EHLO`. I think this could be done by just factoring out the implementation of `EHLO` and adding some configuration-based guards/conditions to the `handle_request` matches for `HELO`, `EHLO`, and `LHLO`.

Second, while an (E)SMTP server returns a single reply after the final `.` of the `DATA` command, an LMTP server returns one reply per recipient address. This is the real point of LMTP: it makes it possible to make a mail delivery server which does not implement a durable queue and full DSN semantics. Since gen_smtp is responsible for generating the protocol response to the completed `DATA` command this would require changes, but perhaps just to use a list in the result tuple when in LMTP mode.

Implementing LMTP in the client would have similar requirements: send `LHLO` instead of `EHLO` and handle the multiple responses to `DATA` and pass them back to the caller. I don't think LMTP client support is as useful as LMTP server support since people are more likely to be implementing some sort of mail-handler that sits behind an MTA than they are to want to write a proper queuing MTA that delivers to LMTP services, but it wouldn't hurt anything either.

I'm not very confident in Erlang (I'm using gen_smtp from Elixir), but I may take a crack at adding LMTP support to the server once I get to that point in my project. If someone else wants to implement it please go ahead.

Shemeikka:
Hi,

I'm interested in LMTP client support. Use case is to receive emails over proprietary protocol and save them to Dovecot using LMTP.

Are maintainers interested in this?

What changes are required to implement LMTP client support? It seems that it could be simple to implement just like @Elemecca wrote. At least Python does it in a [pretty simple way](https://github.com/python/cpython/blob/3.9/Lib/smtplib.py#L1052)

abravalheri:
Hi guys, I gave it a try on #255, but I am not very experienced with the protocol. So if you are willing to have a look, any feedback would be appreciated.

seriyps:
[...]

PR #255, opened 2021-03-22
Add LMTP as an option for the protocol implementation

abravalheri:
Hello,

I would like to thank the maintainers for the amazing work and propose this PR as my attempt to add support for LMTP. This feature was requested in #196, and is something I find very appropriate, mainly because of the following reasons:

1. LMTP is a very similar protocol to SMTP. In practice it looks more of a variation of SMTP than a completely different protocol... Therefore it makes sense to implement both of them together and avoid code duplication between two different projects and code base.
2. Instead of dealing with the complexity of implementing robust and secure SMTP servers by themselves, a considerable part of developers prefer to deploy other battle-tested solution (with advanced features like anti-spam/anti-virus/grey listing integration like Postfix, OpenSMTPD and Exim) and only then place a custom LMTP server behind it.

The implementation I attempted follow more or less the guidelines stablished in #196 and any suggestions for improvement are welcome (in fact this is my very first time writing something in Erlang, so still learning...).

I have also implemented a proof of concept/integration test using OpenSMTPD and docker in https://github.com/abravalheri/lmtp_poc, and things seem to work fine.

Unfortunately I was not able to think in an easy solution for optimising the separation of concerns without requiring duplication of code, but if the maintainers have any ideas and would like to guide me in the process, I am open to go through more iterations of this PR.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -195,6 +195,12 @@ Session options are:
 * `{tls_options, [ssl:server_option()]}` - options to pass to `ssl:handshake/3` (OTP-21+) / `ssl:ssl_accept/3`
   when `STARTTLS` command is sent by the client. Only needed if `STARTTLS` extension
   is enabled
+* `{protocol, smtp | lmtp}` - when `lmtp` is passed, the control flow of the
+  [Local Mail Tranfer Protocol](https://tools.ietf.org/html/rfc2033) is applied.
+  LMTP is derived from SMTP with just a few variations and is used by standard
+  [Mail Transfer Agents (MTA)](https://en.wikipedia.org/wiki/Message_transfer_agent), like Postfix, Exim and OpenSMTPD to
+  send incoming email to local mail-handling applications that usually don't have a delivery queue.
+  The default value of this option is `smtp`.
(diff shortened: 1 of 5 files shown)


seriyps (changes requested):
Thanks! I think it looks quite good, but I have no experience with LMTP. Would really appreciate if you implement the changes I asked for.

seriyps (approved):
Great work! Thanks! Let's see what @mworrell and @arjan think.

abravalheri:
Thank you very much for the review @seriyps.

This is also the first time I am working with LMTP, so I am not that experienced either. What I did was to follow the recommendations of @Elemecca / @Shemeikka on #196 + [RFC2033](https://tools.ietf.org/html/rfc2033). Maybe they have a bit more experience and could also have a look?

To increase my confidence with the code, I did implement [a quick experiment using OpenSMTPD to deliver local emails via LMTP to gen_smtp](https://github.com/abravalheri/lmtp_poc).

seriyps (approved):
Should we merge it @mworrell @arjan ?

mworrell (approved):

seriyps:
Merged! Thanks @abravalheri  @mworrell

PR #258, opened 2021-03-25
Don't crash SMTP client when TLS is optional and we failed to handshake

seriyps:
`send_it` expects that `try_smtp_sessions` returns `{ok, Sock}`, but in "restart without TLS" scenario it would return just `Sock`.

I'm quite surprized that Dialyzed did not complain.

This scenario have not been covered by tests, so, a new test was added.


mworrell:
So much for Dialyzer.... good catch!

mworrell (approved):

PR #261, opened 2021-04-28
Remove extra new line on multipart encoding (closes #260)

### LINKED ISSUE #260: Extra LF breaks DKIM in certain forward scenarios (and mail gets rejected)
Scenario:

gen_smtp (latest) that sends a multipart/alternative email to an upstream mailserver.
Upstream mailserver applies DKIM and then forwards to final destination.

Until here all is good.

If the final destination is an Office365 mailserver that forwards to gmail, email is marked as spam by gmail.

So, in order to recap:
- gen_stmp -> mailserver (dkim) -> Office365 ok
- gen_stmp -> mailserver (dkim) -> Gmail ok
- gen_stmp -> mailserver (dkim) -> Office365 -> Gmail not ok (mail detected as spam)

After investigating and comparing to other mail clients, we discovered that when encoding a multipart body an extra whiteline is added between the `multipart/alternative` component and the parts, for example

```
--_=4g116a4p31245o2q4h634e3y354m3103=_
Content-Type: multipart/alternative;
	boundary="_=513719676i276h6k6a4h1j326j3m5c32=_"
Content-Disposition: inline

<<<< I'm the Extra Empty Line >>>
--_=513719676i276h6k6a4h1j326j3m5c32=_
Content-Type: text/plain;
	 charset="utf-8"
Content-Transfer-Encoding: quoted-printable
Content-Disposition: inline

...omissis...
--_=513719676i276h6k6a4h1j326j3m5c32=_
Content-Type: text/html;
	 charset="utf-8"
Content-Transfer-Encoding: quoted-printable
Content-Disposition: inline

...omissis...
--_=513719676i276h6k6a4h1j326j3m5c32=_--

--_=4g116a4p31245o2q4h634e3y354m3103=_
Content-Transfer-Encoding: base64
Content-Id: <calendar.ics>
Content-Type: text/calendar;
	 method="REQUEST"
Content-Disposition: inline

...omissis...

--_=4g116a4p31245o2q4h634e3y354m3103=_--
```

When Office365 forwards the email, it "cleans up" that empty line, resulting into the DKIM body signature no more valid, so Gmail mark it as spam.

Other clients do not add this extra empty line when composing multipart/mixed content.

The RFC is not very clear if it's allowed or not. Looking into the code the extra line is added [here](https://github.com/gen-smtp/gen_smtp/blob/master/src/mimemail.erl#L888) which should not be needed since `encode_component` [adds it already](https://github.com/gen-smtp/gen_smtp/blob/master/src/mimemail.erl#L869) at the beginning of the component.

A quick test done by removing that extra line seems to fix the issue, but I'm not sure if it is correct or breaks something else.

xadhoom:
An extra new line was added when a multipart/alternative is encoded inside a multipart/mixed email.

This also adds a fixture and a couple of test for both decoding and encoding.

### DIFF
--- a/src/mimemail.erl
+++ b/src/mimemail.erl
@@ -885,7 +885,7 @@ encode_component(_Type, _SubType, Headers, _Params, Body) ->
 
 encode_component_part({<<"multipart">>, SubType, Headers, PartParams, Body}) ->
 	{FixedParams, FixedHeaders} = ensure_content_headers(<<"multipart">>, SubType, PartParams, Headers, Body, false),
-	encode_headers(FixedHeaders) ++ [<<>>] ++
+	encode_headers(FixedHeaders) ++
 	encode_component(<<"multipart">>, SubType, FixedHeaders, FixedParams, Body);
 encode_component_part({Type, SubType, Headers, PartParams, Body}) ->
 	PartData = case Body of
@@ -1528,6 +1528,16 @@ parse_example_mails_test_() ->
 				?assertEqual(?IMAGE_MD5, erlang:md5(element(5, Image)))
 			end
 		},
+		{"alternative text/html with calendar attachment.",
+			fun() ->
+				Decoded = Getmail("message-text-html-attachment.eml"),
+				?assertMatch({<<"multipart">>, <<"mixed">>, _, _, [
+					{<<"multipart">>, <<"alternative">>, _, _, [
+						{<<"text">>, <<"plain">>, _, _, _}, 
+						{<<"text">>, <<"html">>, _, _, _}]}, 
+					{<<"text">>, <<"calendar">>, _, _, _}]}, Decoded)
+			end
+		},
 		{"Outlook 2007 with leading tabs in quoted-printable.",
 			fun() ->
 				Decoded = Getmail("outlook-2007.eml"),
@@ -1950,6 +1960,10 @@ rfc2047_decode_test_() ->
 	].
 
 encoding_test_() ->
+	Getmail = fun(File) ->
+		{ok, Email} = file:read_file(filename:join("test/fixtures/", File)),
+		decode(Email)
+	end,
 	[
 		{"Simple email",
 			fun() ->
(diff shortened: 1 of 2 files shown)


src/mimemail.erl
```diff
@@ -1950,6 +1960,10 @@ rfc2047_decode_test_() ->
 	].
 
 encoding_test_() ->
+	Getmail = fun(File) ->
+		{ok, Email} = file:read_file(string:concat("test/fixtures/", File)),
```
seriyps, line 1964: please use `filename:join` instead of `string:concat`

test/fixtures/message-text-html-attachment.eml
```diff
@@ -0,0 +1,57 @@
...
+Subject: A message with text, html and a calendar attachment
+Reply-To: sender@example.com
+Mime-Version: 1.0
+Sender: sender@example.com
+Content-Type: multipart/mixed;
+	boundary="_=4g116a4p31245o2q4h634e3y354m3103=_"
+Date: Tue, 27 Apr 2021 12:45:20 +0200
+Message-ID: <3c52f2b1dd3d53eff90acf73500fc058@someclient.example.com>
+
+
```
seriyps, line 13: So, it used to add one more newline here?

test/fixtures/message-text-html-attachment.eml
```diff
@@ -0,0 +1,57 @@
...
+--_=4g116a4p31245o2q4h634e3y354m3103=_
+Content-Type: multipart/alternative;
+	boundary="_=513719676i276h6k6a4h1j326j3m5c32=_"
+Content-Disposition: inline
+
+
```
seriyps, line 19: or here?

src/mimemail.erl
```diff
@@ -1528,6 +1528,16 @@ parse_example_mails_test_() ->
 				?assertEqual(?IMAGE_MD5, erlang:md5(element(5, Image)))
 			end
 		},
+		{"alternative text/html with calendar attachment.",
+			fun() ->
+				Decoded = Getmail("message-text-html-attachment.eml"),
+				?assertMatch({<<"multipart">>, <<"mixed">>, _, _, [{
```
seriyps, line 1534: minor: I think it will be a bit easier to read if the last symbol on this line `{` is on the line 1535 `{<<"multipart'>, ...` instead

seriyps (commented):
Thank you! Could you, please, adress the comments and questions? Otherwise looks good!

test/fixtures/message-text-html-attachment.eml
xadhoom: Yes here.

test/fixtures/message-text-html-attachment.eml
xadhoom: No, in the part below, between the `multipart/alternative` component and the it's first part

src/mimemail.erl
xadhoom: fixed in latest push

src/mimemail.erl
xadhoom: fixed

seriyps (approved):

mworrell:
Thanks!

PR #264, opened 2021-06-15
Enable using RSET after transaction failures in deliver/2

juhlig:
The current client implementation of `gen_smtp_client:deliver/2` always quits the connection to the server when it receives a failure reply in answer to `MAIL FROM`, `RCPT TO`, `DATA`, or the mail body.
Instead, `RSET` could be used to reset (ie cancel) the transaction and keep the connection open.

This PR introduces the option `on_transaction_error` which can be used in a `gen_smtp_client:open/1` call. Possible values are `quit` (the default, resulting in the current behavior) and `reset`.

When `reset` is given for this option, the client will try to issue a `RSET` command after receiving a `4xx` or `5xx` response code in answer to `MAIL FROM`, `RCPT TO` or `DATA` (other unexpected replies will cause a quit), and leave the socket connected if the server replies with `250` to the `RSET` command (otherwise it will quit).
If it receives a 4xx or 5xx response after sending the mail body, the transaction is already finished and no `RSET` is needed, but the socket is kept open. Any other unexpected replies will cause a quit.

### DIFF
--- a/src/gen_smtp_client.erl
+++ b/src/gen_smtp_client.erl
@@ -31,7 +31,8 @@
 		{tls_options, [{versions, ['tlsv1', 'tlsv1.1', 'tlsv1.2']}]}, % used in ssl:connect, http://erlang.org/doc/man/ssl.html
 		{auth, if_available},
 		{hostname, smtp_util:guess_FQDN()},
-		{retries, 1} % how many retries per smtp host on temporary failure
+		{retries, 1}, % how many retries per smtp host on temporary failure
+		{on_transaction_error, quit}
 	]).
 
 -define(AUTH_PREFERENCE, [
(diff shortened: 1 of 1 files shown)


seriyps:
Hi, thanks for your PR and sorry for not reviewing it for so long. I think it's a great idea, but I think it would be really nice if you will add a couple of tests where you set this option to `reset` and assert that the behaviour is different.

juhlig:
> Hi, thanks for your PR and sorry for not reviewing it for so long.

No problem :)

> I think it's a great idea, but I think it would be really nice if you will add a couple of tests where you set this option to `reset` and assert that the behaviour is different.

I added two tests, one for RSET, one for QUIT. They are a bit more complicated than the others, the feature can only be tested fully with `deliver` to see that the connection stays open (or gets closed), not with `send` which implicitly closes the connection.

seriyps (approved):

mworrell (approved):

PR #265, opened 2021-06-21
Accept tls options (specifically log_level)

alli5723:
The latest [change to erlang's SSL](https://github.com/erlang/otp/pull/4805/files) module prints a warning when verify_none is used as a verify option. This function only prints at a log level of warning, so it is possible for users to choose when they want to see such warnings.

Accept TLS option for setting ssl_connection log_level.

### DIFF
--- a/src/gen_smtp_client.erl
+++ b/src/gen_smtp_client.erl
@@ -695,7 +695,12 @@ connect(Host, Options) ->
 		undefined -> [];
 		Other -> Other
 	end,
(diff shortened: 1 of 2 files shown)


mworrell:
Just wondering, why not pass the `tls_options` socket options in the `sockopts`?

alli5723:
@mworrell That's a very valid point, my PR is indeed irrelevant with the availability of `sockopts`. 
I missed it, thanks.

mworrell:
Thanks!

PR #266, opened 2021-06-21
Add log_level as SSL_CONNECT_OPTIONS_KEYS

alli5723:
The latest change to erlang's SSL module prints a warning when verify_none is used as a verify option. This function only prints at a log level of warning, so it is possible for users to choose when they want to see such warnings.

This PR simply adds the log_level as one of the acceptable `SSL_CONNECT_OPTIONS_KEYS`, hence users are able to add log_level using sockopts as shown below.

```erlang
...
sockopts: [log_level: :error]
...
```

### DIFF
--- a/src/gen_smtp_client.erl
+++ b/src/gen_smtp_client.erl
@@ -695,7 +695,12 @@ connect(Host, Options) ->
 		undefined -> [];
 		Other -> Other
 	end,
-	SockOpts = [binary, {packet, line}, {keepalive, true}, {active, false} | AddSockOpts],
+	AddTLSOpts = case proplists:get_value(tls_options, Options) of
+		undefined -> [];
+		TLSOpts -> TLSOpts
+	end,
+	AdditionalOpts = [binary, {packet, line}, {keepalive, true}, {active, false} | AddSockOpts],
+	SockOpts = [AdditionalOpts | AddTLSOpts],
 	Proto = case proplists:get_value(ssl, Options) of
 		true ->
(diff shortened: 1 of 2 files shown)


alli5723:
@mworrell and @seriyps, I have recreated the simple PR by adding `log_level` to the available keys `SSL_CONNECT_OPTIONS_KEYS`

seriyps:
Hi, sorry for the late reply. In #274 we disabled any filtering of socket options, so now you should be able to provide whatever TLS and TCP options you like. May you, please, verify that the current master branch works fine for your scenario?

I believe smth like this should do the trick:

```
gen_smtp_client:send(..., [{tls_options, [{log_level, error}]}])
```

(however I personally would recommend to try to start using `verify_peer` as described in https://erlef.github.io/security-wg/secure_coding_and_deployment_hardening/ssl)

alli5723:
> Hi, sorry for the late reply. In #274 we disabled any filtering of socket options, so now you should be able to provide whatever TLS and TCP options you like. May you, please, verify that the current master branch works fine for your scenario?
> 
> I believe smth like this should do the trick:
> 
> ```
> gen_smtp_client:send(..., [{tls_options, [{log_level, error}]}])
> ```
> 
> (however I personally would recommend to try to start using `verify_peer` as described in https://erlef.github.io/security-wg/secure_coding_and_deployment_hardening/ssl)

@seriyps, the latest master satisfies my use-case. 
Thank you and I agree with your recommendation.

PR #272, opened 2021-08-20
mimemail: test binary attachents with proper

seriyps:
Some more property-based tests, testing binary attachments this time (used to be plaintext or UTF-8 only before).


mworrell (approved):

seriyps:
Thanks for the quick review!

PR #273, opened 2021-08-20
Add OTP-24 to CI

seriyps:
Rebar3 is updated to
```
$ ./rebar3 --version
rebar 3.15.2 on Erlang/OTP 24 Erts 12.0
```
(downloaded it from https://github.com/erlang/rebar3/releases/tag/3.15.2)

Have to make some changes in gen_smtp_server_session.erl because on some OTP versions it started to ignore `binary` option after TLS connection upgrade


.github/workflows/ci.yml
```diff
@@ -25,11 +25,27 @@ jobs:
...
+
       - name: Test
         run: make test
+        if: "${{ matrix.otp_vsn != '24.0' }}"
       - name: Proper
         run: make proper
+        if: "${{ matrix.otp_vsn != '24.0' }}"
       - name: Cover
         run: make cover
+        if: "${{ matrix.otp_vsn != '24.0' }}"
```
seriyps, line 47: I didn't manage to find a better way to use `ranch20` profile for OTP-24 and `test` profile for al lother versions. Any ideas?

src/gen_smtp_server_session.erl
```diff
@@ -709,7 +710,7 @@ handle_request({<<"STARTTLS">>, <<>>}, #state{socket = Socket, module = Module,
 			case ranch_ssl:handshake(Socket, [{packet, line}, {mode, list}, {ssl_imp, new} | TlsOpts2], 5000) of %XXX: see smtp_socket:?SSL_LISTEN_OPTIONS
 				{ok, NewSocket} ->
 					?log(debug, "SSL negotiation sucessful~n"),
-					ranch_ssl:setopts(NewSocket, [{packet, line}]),
+					ranch_ssl:setopts(NewSocket, [{packet, line}, binary]),
```
seriyps, line 713: for some reason, without this change `ssl` socket started to send lines as strings instead of binaries on OTP-24

.github/workflows/ci.yml
seriyps: I could probably be more explicit and add rebar profile to the matrix, so we run all the tests with both ranch versions, but then exclude some unsupported combinations (like `ranch 1.7 + OTP-24` `ranch 2.0 + OTP<21.3`). Maybe no need to run dialyzer on all of the though...
smth like
```yaml
matrix:
    otp_vsn: [20.3, 21.3, 22.3, 23.3, 24.0]
    profile: [test, ranch20]
    exclude:
      - otp_vsn: 24.0
        profile: test
      - otp_vsn: 20.3
        profile: ranch20
...
steps:
  - name: Test
    run: make test REBAR_PROFILE=${{matrix.profile}}
```

seriyps:
Hey, @mworrell @arjan . Would you be able to review this when you have time?

cw789 (approved):
I've tried to review now.
Locally testing the server as described in the README.markdown seems to work fine.
But I can only say it for OTP-24.

.github/workflows/ci.yml
```diff
@@ -10,13 +10,19 @@ on:
...
 jobs:
   ci:
-    name: Run checks and tests over ${{matrix.otp_vsn}} and ${{matrix.os}}
+    name: Checks and tests on OTP-${{matrix.otp_vsn}}, ${{matrix.os}}, ${{matrix.profile}}
     runs-on: ${{matrix.os}}
 
     strategy:
       matrix:
-        otp_vsn: [20.3, 21.3, 22.3, 23.1]
+        otp_vsn: [20.3, 21.3, 22.3, 23.3, 24.0]
```
mworrell, line 18: I think we can drop OTP 20, as rebar3 is only supporting the last three OTP releases.

For this reason we have also removed OTP 20 from the testing matrix in Zotonic.

.github/workflows/ci.yml
```diff
@@ -10,13 +10,19 @@ on:
...
         os: [ubuntu-latest]
+        profile: [test, ranch20]
+        exclude:
+            - otp_vsn: 24.0     # ranch 1.x does not work on OTP24+
```
mworrell, line 22: We are using ranch 1.8 with OTP 24.0.5.
I suspect any problems with 1.x and OTP 24 were resolved in 1.8

mworrell:
After this is merged we can rebase #267

src/gen_smtp_server_session.erl
mworrell: That is weird, sounds like a bug in OTP.

arjan: Yes I traced this one down here: https://github.com/erlang/otp/issues/4585

.github/workflows/ci.yml
seriyps: Ah, quite possible 1.8 might work. We just don't have rebar.config profile that enforces 1.8. Do you suggest I should add `ranch18` profile and add it to the matrix as well? Or it's fine as it is now?

.github/workflows/ci.yml
seriyps: `rebar3` version I'm bundling in this PR `3.15.2` works fine for all versions from 20 to 24. But maybe I'm just lucky. I was jsut thinking that #267 would have to drop OTP-20 (because `logger` was introduced in OTP-21), so, it can be done there?

src/gen_smtp_server_session.erl
seriyps: Ah, so we [fixed it for SMTP client](https://github.com/gen-smtp/gen_smtp/commit/f157a572d3ea11bc223c0e72c3e037532842c6c6), but forgot about the server :smile:
But good to know it's not some mystery but actual known OTP issue.

seriyps:
ok, so I removed OTP-20 and also added one more test profile that includes ranch 1.8

mworrell:
Looks good!

PR #277, opened 2021-09-12
Switch to SASL subprotocol on AUTH command

seriyps:
I noticed that Python's [SMTPLIB](https://docs.python.org/3/library/smtplib.html) sends all verbs in lowercase and gen_smtp fails to recognize `rset\r\n` command.

UPD: while digging deeper, I found https://datatracker.ietf.org/doc/html/rfc4954 and it says that `AUTH xxxx` verb actually makes client/server switch to a SASL subprotocol which uses base64-encoded data exchange. So, it does not make sense to use the same `handle_request`/`parse_request` functions as the ones we use for regular SMTP.
RFC also says that client may interrupt SASL subprotocol by sending `*\r\n`, but I'm not implementing this feature here.


mworrell (approved):

mworrell:
That is quite a change (and a good one) for something that looked so simple.

seriyps:
The change itself is not that big, mostly moved  some clauses of `handle_request` to their own function. But yes, quite surpizing!

PR #278, opened 2021-09-13
Replace RFC-822 address-list parser with RFC-5322

seriyps:
A more modern and robust parser for address specifiers like this (may appear in To / From / CC etc):
https://datatracker.ietf.org/doc/html/rfc5322#section-3.4

-  `login@domain`
- `Name <login@domain>`
- `Name Surname <login@domain>`
- `Name <login1@domain1>, Name2 <login2@domain2>`
- `group name:login@domain,Name <login2@domain2>;`

 Also different versions of escaping of name / login / domain.
It's much more unicode-friendly also.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,8 @@ build
 .eunit/
 ebin/gen_smtp.app
 src/smtp_rfc822_parse.erl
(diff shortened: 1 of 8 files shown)


mworrell:
Great work!

PR #279, opened 2021-09-13
Use SPDX ID for licenses

Cantido:
Hex.pm recommends this, and it helps tools identify exactly which BSD license this project is released under.


seriyps:
Thanks! Didn't know it matters.

seriyps (approved):

Cantido:
It's a little thing, but the package info is available via the hex API so it just makes it a little easier for tools to detect your license. I'm actually writing a tool that checks dependency licenses, and uses these IDs to make sure all the dependencies I'm using have recognized licenses.

PR #280, opened 2021-09-16
Add support for SMTPUTF8 extension, RFC-6531

seriyps:
https://datatracker.ietf.org/doc/html/rfc6531
It's relatively simple extension that allows SMTP clients send `MAIL FROM: <utf-8 email>` and `RCPT TO: <utf-8 email>`.
I never seen UTF-8 email addresses in a wild so far, but I saw many SMTP servers implementing this extension (including gmail and outlook).

### DIFF
--- a/src/gen_smtp_server_session.erl
+++ b/src/gen_smtp_server_session.erl
@@ -35,7 +35,7 @@
 -endif.
 
 -define(DEFAULT_MAXSIZE, 10485760). %10mb
--define(BUILTIN_EXTENSIONS, [{"SIZE", integer_to_list(?DEFAULT_MAXSIZE)}, {"8BITMIME", true}, {"PIPELINING", true}]).
+-define(BUILTIN_EXTENSIONS, [{"SIZE", integer_to_list(?DEFAULT_MAXSIZE)}, {"8BITMIME", true}, {"PIPELINING", true}, {"SMTPUTF8", true}]).
 -define(TIMEOUT, 180000). % 3 minutes
 
 %% External API
@@ -55,7 +55,8 @@
 		to = [] :: [binary()],
 		data = <<>> :: binary(),
 		expectedsize = 0 :: pos_integer() | 0,
-		auth = {<<>>, <<>>} :: {binary(), binary()} % {"username", "password"}
+		auth = {<<>>, <<>>} :: {binary(), binary()}, % {"username", "password"}
+		flags = [] :: [smtputf8 | '8bitmime']
 	}
 ).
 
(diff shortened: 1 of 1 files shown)


src/gen_smtp_server_session.erl
```diff
@@ -546,7 +554,7 @@ handle_request({<<"MAIL">>, Args},
 									?log(debug, "error: ~s~n", [Message]),
 									send(State, Message),
 									{ok, State};
-								NewState ->
+								#state{envelope = Envelope} = NewState ->
```
seriyps, line 557: it's actually a small bugfix, because we used to loose the value of `#envelope.expectedsize` we set on lines 522 and 528

src/gen_smtp_server_session.erl
```diff
@@ -756,62 +764,73 @@ handle_error(Kind, Details, #state{module=Module, callbackstate = OldCallbackSta
 			State
 	end.
 
--spec parse_encoded_address(Address :: binary()) -> {binary(), binary()} | 'error'.
-parse_encoded_address(<<>>) ->
```
seriyps, line 760: most of the diff in `parse_encoded_address` is just me replacing the tuple with the record (except one clause I added and changing `list_to_binary` to `unicode:characters_to_binary`)

src/gen_smtp_server_session.erl
```diff
@@ -756,62 +764,73 @@ handle_error(Kind, Details, #state{module=Module, callbackstate = OldCallbackSta
...
-	parse_encoded_address(Tail, [H | Acc], {false, AB}); % digits
-parse_encoded_address(<<H, Tail/binary>>, Acc, {false, AB}) when H >= $@, H =< $Z ->
-	parse_encoded_address(Tail, [H | Acc], {false, AB}); % @ symbol and uppercase letters
-parse_encoded_address(<<H, Tail/binary>>, Acc, {false, AB}) when H >= $a, H =< $z ->
-	parse_encoded_address(Tail, [H | Acc], {false, AB}); % lowercase letters
-parse_encoded_address(<<H, Tail/binary>>, Acc, {false, AB}) when H =:= $-; H =:= $.; H =:= $_ ->
-	parse_encoded_address(Tail, [H | Acc], {false, AB}); % dash, dot, underscore
+parse_encoded_address(<<H/utf8, Tail/binary>>, Acc, #pa{utf8 = true} = F) when H > 127 ->
+	%% https://datatracker.ietf.org/doc/html/rfc6531#section-3.3
+	parse_encoded_address(Tail, [H | Acc], F); % UTF-8 above 7bit (when allowed)
```
seriyps, line 816: this is the only new clause I added

src/gen_smtp_server_session.erl
```diff
@@ -1028,66 +1047,83 @@ parse_encoded_address_test_() ->
...
-					?assertEqual({<<"God~*$@heaven.af.mil">>, <<>>}, parse_encoded_address(<<"<God~*$@heaven.af.mil>">>))
+					?assertEqual({<<"God@heaven.af.mil">>, <<>>}, parse_encoded_address(<<"<God@heaven.af.mil>">>, false)),
+					?assertEqual({<<"God@heaven.af.mil">>, <<>>}, parse_encoded_address(<<"<\\God@heaven.af.mil>">>, false)),
+					?assertEqual({<<"God@heaven.af.mil">>, <<>>}, parse_encoded_address(<<"<\"God\"@heaven.af.mil>">>, false)),
+					?assertEqual({<<"God@heaven.af.mil">>, <<>>}, parse_encoded_address(<<"<@gateway.af.mil,@uucp.local:\"\\G\\o\\d\"@heaven.af.mil>">>, false)),
+					?assertEqual({<<"God2@heaven.af.mil">>, <<>>}, parse_encoded_address(<<"<God2@heaven.af.mil>">>, false)),
+					?assertEqual({<<"God+extension@heaven.af.mil">>, <<>>}, parse_encoded_address(<<"<God+extension@heaven.af.mil>">>, false)),
+					?assertEqual({<<"God~*$@heaven.af.mil">>, <<>>}, parse_encoded_address(<<"<God~*$@heaven.af.mil>">>, false)),
+					?assertEqual({<<"God~!#$%^&*()_+123@heaven.af.mil">>, <<>>}, parse_encoded_address(<<"<\"God~!#$%^&*()_+123\"@heaven.af.mil>">>, false))
```
seriyps, line 1057: the only extra test in "Valid addresses should parse" section, and it is not related to the PR's topic. I just noticed that quoted address is not covered by tests

mworrell:
I agree, supporting this is good, even if we didn't see it in the wild yet.

PR #283, opened 2021-09-22
bugfix: when 8BITMIME extension is enabled, BODY could be "7bit"

seriyps:
I unfortunately introduced a bug in https://github.com/gen-smtp/gen_smtp/commit/ecc923e39f2d11af6977ccf0e3238ab6e8c754d7

Turns out, many SMTP senders are sending `MAIL FROM: .... BODY=7bit` when they see `8BITMIME` extension.
see https://datatracker.ietf.org/doc/html/rfc6152#section-2


mworrell:
Mailers are funny.  Good catch.

PR #286, opened 2021-09-29
Minor bugfixes and improvements

juhlig:
Will comment inline :)

### DIFF
--- a/src/gen_smtp_server_session.erl
+++ b/src/gen_smtp_server_session.erl
@@ -401,25 +401,20 @@ handle_request({Msg, Hostname},
 			end,
 			Extensions2 = case Tls of
 				true ->
-					Extensions1 -- [ {"STARTTLS", true} ];
+					lists:delete({"STARTTLS", true}, Extensions1);
 				false ->
 					Extensions1
 			end,
-			{_, _, Response} = lists:foldl(
-				fun
-					({E, true}, {Pos, Len, Acc}) when Pos =:= Len ->
-						{Pos, Len, [["250 ", E, "\r\n"] | Acc]};
-					({E, Value}, {Pos, Len, Acc}) when Pos =:= Len ->
-						{Pos, Len, [["250 ", E, " ", Value, "\r\n"] | Acc]};
-					({E, true}, {Pos, Len, Acc}) ->
-						{Pos+1, Len, [["250-", E, "\r\n"] | Acc]};
-					({E, Value}, {Pos, Len, Acc}) ->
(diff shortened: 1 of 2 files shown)


src/gen_smtp_server_session.erl
```diff
@@ -401,25 +401,20 @@ handle_request({Msg, Hostname},
...
-				{1, length(Extensions2), [ ["250-", hostname(Options), "\r\n"] ]},
-				Extensions2),
+			Response = (fun
+				F([{E, true}]) -> ["250 ", E, "\r\n"];
+				F([{E, V}]) -> ["250 ", E, " ", V, "\r\n"];
+				F([Line]) -> ["250 ", Line, "\r\n"];
+				F([{E, true}|More]) -> ["250-", E, "\r\n" | F(More)];
+				F([{E, V}|More]) -> ["250-", E, " ", V, "\r\n" | F(More)];
+				F([Line|More]) -> ["250-", Line, "\r\n" | F(More)]
+			end)([hostname(Options)|Extensions2]),
```
juhlig, line 415: This is more elegant IMO, it doesn't have to carry around the Pos and Len to decide if it is the last line. Also, with the old solution, it might have happened that `Extensions2` was empty (depending on circumstances, "SIZE" and/or "STARTTLS" could be removed above), but the hostname line would still be created with a dash instead of a space after the code.

I could have put this in an own function, but it is the only place in the module where a multiline response is created, so...

src/gen_smtp_server_session.erl
```diff
@@ -401,25 +401,20 @@ handle_request({Msg, Hostname},
 			end,
 			Extensions2 = case Tls of
 				true ->
-					Extensions1 -- [ {"STARTTLS", true} ];
+					lists:delete({"STARTTLS", true}, Extensions1);
```
juhlig, line 404: `lists:delete` is faster than `--`, and more intuitive here IMO.

src/smtp_util.erl
```diff
@@ -56,7 +56,7 @@ mxlookup(Domain) ->
 		[] ->
 			lists:map(fun(X) -> {10, inet_parse:ntoa(X)} end, inet_res:lookup(Domain, in, a));
 		Result ->
-			lists:sort(fun({Pref, _Name}, {Pref2, _Name2}) -> Pref =< Pref2 end, Result)
+			lists:sort(Result)
```
juhlig, line 59: There is no need for the sorting function, natural sorting will give the same effective result. A minor difference is that entries with the same preference will also be sorted by `Name`s, which is of no consequence.

seriyps (approved):
Cool, thanks!

PR #287, opened 2021-09-29
Update .editorconfig

cw789:
Probably bikeshedding.

But with tab the `indent_size` should be an user preference.
Additionally if added `.md` for possibly future guides with this extension.


mworrell:
I am not sure about the tab change.

We might have mixed source files, and sometimes a combination of spaces and tabs is used for alignment.

cw789:
Your're right. With the existing mixed source files it's better to keep the pre configured `ident_size` for the tabs.
Otherwise it makes the code less readable for users with different configured tab size (provided they use an .editorconfig compatible editor - otherwise it's anyway disordered).

So I've reverted the change.

It's now only resorting some things in the file.
Adding `.md` as extension (for future guides I possibly will write).
Removing `.conf` as I don't find an extension like that in the repository.

So close or merge. I'm happy with both.

seriyps (approved):
Yep, unfortunately the coding style we have in gen_smtp is quite special =)
It would be cool to add some code-formatter like https://github.com/WhatsApp/erlfmt one day, but the problem with it is that it messes-up with the `git blame`...

cw789:
Ok, I did one last quick edit.
I changed the `ident_size` to 2 for `.yml` as it is currently in there.

mworrell:
@cw789 is this ready for merge?

cw789:
> @cw789 is this ready for merge?

Yes it is now.

mworrell:
Thanks!

cw789:
> Yep, unfortunately the coding style we have in gen_smtp is quite special =)
> It would be cool to add some code-formatter like https://github.com/WhatsApp/erlfmt one day, but the problem with it is that it messes-up with the `git blame`...

Possibly sometime an evaluation worth. But the thing with git blame is a good point. Wouldn't have thought about that.

PR #288, opened 2021-09-30
CI refactoring

cw789:
Refactoring CI to not be in need to ship the rebar3 binary within the repository itself.
Additionally it newly includes caching.
The tests now run on every push and pull (for now?).
OTP 21 uses rebar3 3.15.2, newer OTP the currently latest rebar3 3.17.0.

Matrix:
I would build EDoc only on latest OTP. I don't get the benefit of building it on various versions (but you could proof me wrong).
We could expand tests to run on more OTP / rebar3 combinations? I don't know if needed.


cw789:
Ready for review or discussions now.

.github/workflows/ci.yml
```diff
@@ -1,39 +1,56 @@
...
+          otp-version: ${{ matrix.otp }}
+          rebar3-version: ${{ matrix.rebar3 }}
+
+      - uses: actions/cache@v2
+        with:
+          path: |
+            ~/.cache/rebar3
+            _build
+          key: ${{ runner.os }}-erlang-${{ matrix.otp }}-${{ hashFiles('**/rebar.lock') }}
+          restore-keys: ${{ runner.os }}-erlang-${{ matrix.otp }}-
```
seriyps, line 41: I never used `acgions/cache` before... Is it safe to cache `_build` directory?

seriyps (approved):
Thanks!

.github/workflows/ci.yml
cw789: That is a very good question. At least it is save with Elixir Github Action workflows.
I can't say for sure how tolerant Erlang is with this behavior.

cw789: Yes it seem so that compilation is quick enough. So the benefit of caching whole `_build` is negligible.

Only dialyzer is slow and benefits a lot from caching the `plt`.
I would suggest to write a rule that contains the plt within `_build/dialyzer` only.

`.cache/rebar3` seems to be fine - not bringing much benefit - but not doing any harm either.

seriyps: I think `.cache/rebar3` is actually nice because it holds hex.pm cache (eg ranch) as well as dialyzer .plt for OTP standard library

seriyps: > I would suggest to write a rule that contains the plt within _build/dialyzer only.

Sounds like a good idea

cw789: Done. I've narrowed done the cache specification.
- Hex packages
- Dialyzer PLTs

For now not caching any build files.
It would save some little seconds, but it's not worth it now until we have more experience with it.
(rebar3 is currently still determing recompilation by timestamps)

For the Dialyzer PLTs cache I would have preferred to have the full OTP version in the key.
But as PLTs are not that big of size and restoring cache is blazing fast I could probably forget about it.

cw789:
I did some wrong merge things will tidy up the commits later.

.github/workflows/ci.yml
```diff
@@ -1,39 +1,64 @@
...
+          path: ~/.cache/rebar3/hex/hexpm/packages
+          key: ${{ runner.os }}-hex-${{ hashFiles('**/rebar.lock') }}
+          restore-keys: ${{ runner.os }}-hex-
+
+      - name: Cache Dialyzer PLTs
+        uses: actions/cache@v2
+        with:
+          path: |
+            ~/.cache/rebar3/rebar3_*.plt
+            _build/dialyzer/rebar3_*.plt
```
seriyps, line 47: is it really `.plt`? On my computer it creates `_plt`…

seriyps (approved):
Thanks!

.github/workflows/ci.yml
cw789: No it's is indeed not. And I was thinking about why a second CI run isn't faster. Thanks!

mworrell (approved):

cw789:
Thanks for merging. Doc seems to not run on `master` (default branch) but should to do so.
I'll check that and send an additional PR.

mworrell:
Ah, missed that part reviewing the changes.
Thanks!

PR #290, opened 2021-10-04
Building doc with ExDoc instead of EDoc

cw789:
This is a proof how to build the doc with [ExDoc](https://github.com/elixir-lang/ex_doc) instead of EDoc.
In my personal opinion the UX is better. But probably I'm only more used to it.
This is possible since Erlang/OTP 24+.

You can have a look in the doc action build artifacts to get yourself an impression.
→ https://github.com/gen-smtp/gen_smtp/actions/runs/1320190497

Things to discuss:
- Should we document somewhere for maintainers how to upload the doc to hex?
  -> `rebar3 ex_doc`

Somehow solved:
- I needed to rename the README to `.md`.
  If that's unwanted we can probable make an temp copy within `docs.sh`.
  Edit: I don't find anything negative this change would bring (`git blame` also seems to work as expected).
- ~I'm not totally happy with setting the source ref.~
  ~Even with tags this would currently always only have the commit sha in it.~
  Edit: Probably docs for releasing on hex would anyway be done locally.
- ~I'm unhappy to have an additional place where pumping the `version` is needed.~
  ~Edit: Solved. But not very simple.~
  Edit: Not needed anymore with using `rebar3_ex_doc`.
- ~I do not like that ex_doc install step is currently doing too much work every time.~
  ~I also would prefer if this action wouldn't need Elixir for installing ex_doc via `mix escript.install hex ex_doc --force`.~
 ~Edit: A rebar3 integration for it is on the erlef roadmap...~
  Edit: rebar3 plugin `rebar3_ex_doc` is available and integrated.


src/gen_smtp_server_session.erl
```diff
@@ -150,7 +141,6 @@
 -optional_callbacks([handle_info/2, handle_AUTH/4, handle_error/3]).
 
 %% @doc Start a SMTP session linked to the calling process.
-%% @see start/3
```
cw789, line 153: I didn't find `start/3` being there anymore.

src/gen_smtp_server_session.erl
```diff
@@ -79,16 +79,7 @@
...
-%% OTP-19: ssl:ssl_option()
-%% OTP-20: ssl:ssl_option()
-%% OTP-21: ssl:tls_server_option()
-%% OTP-22: ssl:tls_server_option()
-%% OTP-23: ssl:tls_server_option()
--ifdef(OTP_RELEASE).
 -type tls_opt() :: ssl:tls_server_option().
--else.
--type tls_opt() :: ssl:ssl_option().
--endif.
```
cw789, line 91: It's probably save to remove this conditional as testing doesn't happen for Erlang/OTP < 21 anymore.

.github/workflows/docs.yml
```diff
@@ -35,5 +37,15 @@ jobs:
...
 
-      - name: Build EDoc
-        run: make docs
+      - name: Install ExDoc
+        run: mix escript.install hex ex_doc --force
+
+      - name: Build ExDoc
+        run: sh docs.sh $GITHUB_SHA
+
+      - name: 'Upload artifact'
```
cw789, line 46: Note to myself: not consistant to use '

docs.sh
```diff
@@ -0,0 +1,20 @@
...
+
+version=1.2.0-dev
+
+rebar3 compile
+rebar3 as docs edoc
+
+~/.mix/escripts/ex_doc "gen_smtp" $version "_build/default/lib/gen_smtp/ebin" \
+  --source-ref $SOURCE_REF \
+  --config docs.config \
+  --paths "_build/default/lib/*/ebin"
```
cw789, line 20: Note to myself: addcoutput of installed `ex_doc --version`

docs.sh
```diff
@@ -0,0 +1,20 @@
...
+
+# Usage:
+# ./docs.sh <source-ref>
+
+SOURCE_REF=$1
+
+version=1.2.0-dev
+
+rebar3 compile
+rebar3 as docs edoc
```
seriyps, line 15: so it builds both `edoc` and `ExDoc`? Or is it just to produce documentation chunks?
Because what I was actually about to propose is that we test that we still able to build both Erlang edocs and elixir ExDoc, even if we are still going to publish only one of them.

seriyps (commented):
One difference that I noticed is that `ExDoc` includes non-exported functions to the HTML, while `edoc` seems to only include the functions which are listed in `-export([...]).`. Not sure itmakes sense to include internal functions to docs; especially for the generated modules like `smtp_rfc5322_parse`.

But the looks are really nice!

docs.sh
cw789: Sadly not. The change in `rebar.confic` instructs `edoc` to generate doc chunks instead of HTML.
I would also prefer to build both as a starting point.

I've not checked if `rebar3 edoc` accepts this options somehow as command line parameters?

seriyps: Ah, ok. Well, I don't have any strong opinion on this topic.
I guess it should be possible to re-define the `edoc_opts` by creating some additional profile in `rebar.config`, but I'm not sure if it worth it.

> I've not checked if rebar3 edoc accepts this options somehow as command line parameters?

It seems `edoc` command doesn't take any parameters: https://github.com/erlang/rebar3/blob/ba50a0828eba37d386c5f82cf1814d4519d66f64/src/rebar_prv_edoc.erl#L28

```
$ rebar3 help edoc
===> Analyzing applications...
Generate documentation using edoc.
Usage: rebar3 edoc
```

arjan:
This looks very good imho :-) For some reason exdocs is way more readable than edocs. I'm :+1:

docs.sh
cw789: Fine. So let us then keep this consideration at side for now.
Locally it can be anyway done be commenting out the `edoc_opts` in the `rebar.config` and build doc wit EDoc as usual.

cw789:
> One difference that I noticed is that `ExDoc` includes non-exported functions to the HTML, while `edoc` seems to only include the functions which are listed in `-export([...]).`, see https://hexdocs.pm/gen_smtp/1.1.1/. Not sure itmakes sense to include internal functions to docs; especially for the generated modules like `smtp_rfc5322_parse`.

That's something I definitely need to check. I don't think it makes sense to include private functions.
I also don't think it is supposed to do that (there are some upstream issues about similar things).

To be honest, I haven't compared the two results yet. I just wanted to proof how it could be done first.
Possibly it is still early to use ExDoc with Erlang (https://github.com/elixir-lang/ex_doc/issues/1333).

cw789:
> But the styles are really nice!

> This looks very good imho :-) For some reason exdocs is way more readable than edocs. I'm 👍

Perfect. Then I'll continue to polish things out.

docs.sh
```diff
@@ -0,0 +1,38 @@
...
+
+app_name="gen_smtp"
+
+escript_get_app_vsn=$(cat <<EOF
+{ok, [{_, _, App}]} = file:consult("src/${app_name}.app.src"),
+io:format(proplists:get_value(vsn, App)),
+halt(0)
+EOF
+)
+app_vsn=$(erl -noshell -eval "$escript_get_app_vsn")
```
cw789, line 23: Isn't there any simpler solution to get the app version?

cw789:
So I've revisited this.

I currently can't manage to exclude `smtp_rfc5322_parse` & `smtp_rfc5322_scan` without manually adding `%% @private` to the generated `.erl` files.
Other then that I think we are nearly there.

cw789:
Ok, from my side this is ready. But only if library publishers are OK with it.  

- For now this means if you want to generate the docs locally you'll need Elixir
- EDoc will only emit the chunks with this change
- Generated `smtp_rfc5322_parse` & `smtp_rfc5322_scan` will be included in the docs now

I'm also OK if this PR was just an exploration.
If it'll be revisited when things are more mature or `rebar3` will have `ExDoc` capability baked in.

seriyps (approved):
I personally like it a lot!

cw789:
The rebar3 plugin for generating docs with ExDoc is ready: https://hex.pm/packages/rebar3_ex_doc

I've changed the implementation which simplifies a lot.
Generating docs is now simply done by: `rebar3 ex_doc`
No install of Elixir necessary anymore.

There are some warnings which I haven't explored yet:
```
===> Running edoc for gen_smtp
edoc: warning: 'tt' is not allowed - skipping tag, extracting content
===> Running ex_doc for gen_smtp
===> warning: documentation references type "ranch:opts/0" but it is undefined or private
  src/gen_smtp_server.erl:39: t:gen_smtp_server.options/0

warning: documentation references type "ranch:ref/0" but it is undefined or private
  src/gen_smtp_server_session.erl:147: gen_smtp_server_session.start_link/3

warning: ExDoc.Markdown.Earmark (deprecated) README.md:0 The timeout option has no effect anymore and will be removed in EarmarkParser 1.5

warning: documentation references type "ranch:opts/0" but it is undefined or private
  src/gen_smtp_server.erl:39: t:gen_smtp_server.options/0

warning: documentation references type "ranch:ref/0" but it is undefined or private
  src/gen_smtp_server_session.erl:147: gen_smtp_server_session.start_link/3

warning: ExDoc.Markdown.Earmark (deprecated) README.md:0 The timeout option has no effect anymore and will be removed in EarmarkParser 1.5
```

PS: The failing CI seems to be unrelated to this change.

PR #291, opened 2021-10-05
Fix gen_smtp_client:send/3 with callback

juhlig:
The current implementation of `gen_smtp_client:send/3` with a callback function has some issues:
* different from what the spec says and other than `send/3` with no callback, it returns only a pid, not `{ok, Pid}`
* other than `send/3` with no callback, the spawned (middleman) process is not linked to the calling process, so if the calling process dies, the send operation will be aborted when no callback is used, but left running if a callback is used
* if the `send_it` call in `send_it_nonblock`, called from the inner process causes that process to exit, the outer process calls the Callback with `{exit, Reason}` as argument. However, the same will happen if the `send_it` call succeeds, the result gets passed to the Callback, and then the Callback itself causes an exit of the inner process, so the callback will be called again with `{exit, Reason}` in the outer process
* there are no tests

By the changes in this PR,
* `send/3` will always return `{ok, Pid}`
* the middleman process is removed, and the spawned process is always linked to the calling process
* exits as a consequence of the `send_it` call (which are then passed as `{exit, Reason}` to the callback if one is given, or cause the process to exit if no callback is given) are separated from exits resulting from the Callback (which always cause the process to exit)
* a test case was added

### DIFF
--- a/src/gen_smtp_client.erl
+++ b/src/gen_smtp_client.erl
@@ -137,26 +137,24 @@ send(Email, Options, Callback) ->
 	NewOptions = lists:ukeymerge(1, lists:sort(Options),
 		lists:sort(?DEFAULT_OPTIONS)),
 	case check_options(NewOptions) of
-		ok when is_function(Callback) ->
-			spawn(fun() ->
-						process_flag(trap_exit, true),
-						Pid = spawn_link(fun() ->
-									send_it_nonblock(Email, NewOptions, Callback)
-							end
-						),
-						receive
-							{'EXIT', Pid, Reason} ->
-								case Reason of
-									X when X == normal; X == shutdown ->
-										ok;
-									Error ->
-										Callback({exit, Error})
-								end
-						end
-				end);
 		ok ->
-			Pid = spawn_link(fun () ->
-						send_it_nonblock(Email, NewOptions, Callback)
+			Pid = spawn_link(
+				fun () ->
+					try
+						send_it(Email, NewOptions)
+					of
+						{error, _Type, _Reason} = Error when is_function(Callback, 1) ->
+							Callback(Error);
+						{error, _Type, _Reason} = Error ->
+							exit(Error);
+						Receipt when is_function(Callback, 1) ->
+							Callback({ok, Receipt});
+						_Receipt ->
+							ok
+					catch
+						exit:Reason when is_function(Callback, 1) ->
+							Callback({exit, Reason})
+					end
 				end
 			),
 			{ok, Pid};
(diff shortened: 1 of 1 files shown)


src/gen_smtp_client.erl
```diff
@@ -137,26 +137,24 @@ send(Email, Options, Callback) ->
...
-										ok;
-									Error ->
-										Callback({exit, Error})
-								end
-						end
-				end);
 		ok ->
-			Pid = spawn_link(fun () ->
-						send_it_nonblock(Email, NewOptions, Callback)
+			Pid = spawn_link(
```
seriyps, line 141: since we are already here, maybe use `proc_lib:spawn_link` for it?

src/gen_smtp_client.erl
```diff
@@ -1166,6 +1147,33 @@ session_start_test_() ->
...
+							?assertMatch({ok, "DATA\r\n"}, smtp_socket:recv(X, 0, 1000)),
+							smtp_socket:send(X, "354 ok\r\n"),
+							?assertMatch({ok, "hello world\r\n"}, smtp_socket:recv(X, 0, 1000)),
+							?assertMatch({ok, ".\r\n"}, smtp_socket:recv(X, 0, 1000)),
+							smtp_socket:send(X, "250 ok\r\n"),
+							?assertMatch({ok, "QUIT\r\n"}, smtp_socket:recv(X, 0, 1000)),
+							?assertMatch({ok, <<"ok\r\n">>}, receive {callback, Ref, CbRet} -> CbRet end),
+							ok
+						end
+					}
```
seriyps, line 1175: might be good to have a test that tests failure scenario, but I won't insist

seriyps (approved):
Nice cleanup! Thanks!

src/gen_smtp_client.erl
juhlig: I could, but why? =^^=
I mean, `proc_lib:spawn_link` is just `erlang:spawn_link` with a cherry on top, to make it fit and play nicely in supervision trees or as a special process, neither of which this is?

src/gen_smtp_client.erl
juhlig: I'll see to it later.

juhlig: Done :)

seriyps: Thanks 👍

src/gen_smtp_client.erl
seriyps: I don't have a strong opinion on this. Just have my rule of thumb to use `proc_lib` instead of raw spawns without much thinking :) But let's see what @mworrell or @arjan would say

mworrell:
@seriyps You reviewed this before, is it ready for merge?

seriyps:
@mworrell I think @juhlig only pushed some extra tests for "failure scenarios" after I approved. So, LGTM.

mworrell:
I am not sure I have access to the hex repo, or was the hex push automated?

seriyps:
@mworrell nope, it is not automated. I see only 3 users have admin permissions on hex:

    sleepy_keita
    arjan
    vagabond

PR #292, opened 2021-10-11
Improve Quoted-Printable encoding

Maria-12648430:
The current implementation of `encode_quoted_printable/3` is a bit crude (no offense, really 😜), especially when it comes to lines reaching the limit of 76 characters, resorting to `string` functions to retrieve the last line and finding whitespaces in there.

The improved version `encode_quoted_printable/6` which I'm proposing here works by remembering the occurences of linear whitespaces (including HTABs), and otherwise only look-aheads in the original data to see if a CRLF follows the current character. I also refactored the encoding of characters itself to work via integer operations only, thereby getting rid of all the `io:format` and `lists:flatten` calls.

I admit that my implementation isn't exactly easy on the eye, but that is because QP-encoding itself is tricky. And IMO, the current implementation isn't that easy to figure out, either 😉

I left the current tests untouched in what they test, only replaced the calls to the helper function `encode_quoted_printable/3` (of which I don't see the point) with calls to the API function `encode_quoted_printable/1`. I also removed one test, `"newline craziness"`, of which I also fail to see the point 🤔

Finally, I put in a micro-optimization in `choose_transformation/1` which avoids division and rounding and only uses multiplication to figure out if the percentage of printable characters in the sample is above 80.

[EDIT]: Just noticed an edge case in which a line could exceed the length limit :laughing: Will fix and update the PR tomorrow.

### DIFF
--- a/src/mimemail.erl
+++ b/src/mimemail.erl
@@ -749,15 +749,15 @@ choose_transformation(Body) ->
 	Size = byte_size(Body),
 	% get only the allowed ascii characters
 	% TODO - this might not be the complete list
-	FilteredSize = length([X || <<X>> <= Body, ((X > 31 andalso X < 127) orelse X == $\r orelse X == $\n)]),
-
-	Percent = round((FilteredSize / Size) * 100),
+	FilteredSize = byte_size(<< <<X>> || <<X>> <= Body, ((X > 31 andalso X < 127) orelse X == $\r orelse X == $\n)>>),
 
 	%based on the % of printable characters, choose an encoding
 	if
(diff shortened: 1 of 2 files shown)


Maria-12648430:
Last commit fixes the line length bug I had in my first implementation, adds some tests for a few edge cases, augments another test to ensure that whitespaces are encoded when last character, and removes the trimming from the proptest.

seriyps (approved):
Great, thanks!

mworrell:
I am very happy with this.
Tests are also looking good.

@seriyps shall we merge?

PR #304, opened 2022-01-25
Mimemail parameters fixes

seriyps:
Noticed that in some specific case `mimemail:mimetuple()`'s 4th element which is supposed to always be `mimemail:parameters()` map, it is actually returned as a list (of incorrect values anyway) from `decode/1`.

Fixed it, also added some more proper-tests validation for `parameters()`.

Small bonus: some optimization in `mimemail:get_header_value/2,3`

### DIFF
--- a/src/mimemail.erl
+++ b/src/mimemail.erl
@@ -85,7 +85,8 @@
 -type mime_subtype() :: binary().               % `<<"plain">>'
 -type headers() :: [{binary(), binary()}].      % `[{<<"Content-Type">>, <<"text/plain">>}]'
 -type parameters() ::
-        #{transfer_encoding => binary(),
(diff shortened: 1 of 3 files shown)


src/mimemail.erl
```diff
@@ -143,9 +144,19 @@ decode(OrigHeaders, Body, Options) ->
 					decode_component(Headers, Body, MimeVersion, Options);
 				{<<"multipart">>, _SubType, _Parameters} ->
 					erlang:error(non_mime_multipart);
-				{Type, SubType, Parameters} ->
+				{Type, SubType, CTParameters} ->
```
seriyps, line 147: the bug was here. We were returning the value of "content-type parameters" in place where we supposed to return `mimemail:parameters()` map.

mworrell (approved):

mworrell:
Good work, thanks again! 👍

PR #308, opened 2022-02-12
Generating docs with ExDoc instead of EDoc

cw789:
Clean new PR replacing https://github.com/gen-smtp/gen_smtp/pull/290.

---

Additionally this PR removes code meant for < OTP 21.
Additionally this PR introduces format checking in CI.
This part is probably opinionated if task should be allowed to fail
until every maintainer is used to check the format beforehand.

I've added the usage of `ex_doc` & `fmt` to the Makefile as a kind of documentation.
For myself it is not 100% clear how I could make sure I use the latest rebar3 plugin as they are not anywhere pinned.
-> [Upgrading Plugins](https://rebar3.readme.io/docs/using-available-plugins#upgrading-plugins)
This also means I'm not sure how reproducible the CI is regarding the plugins.

All the best.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,4 +1,4 @@
-name: Doc
+name: Docs
 
 on:
   push:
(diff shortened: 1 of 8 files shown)


seriyps:
I think plugins can be pinned same way as dependencies:

```erlang
{project_plugins, [
    {erlfmt, "1.0.0"},
    {rebar3_proper, "0.12.1"}
]}.
```

Don't know why I haven't done it initially…

seriyps (approved):
Thanks!

.github/workflows/docs.yml
```diff
@@ -35,5 +35,5 @@ jobs:
           key: ${{ runner.os }}-hex-${{ hashFiles('**/rebar.lock') }}
           restore-keys: ${{ runner.os }}-hex-
 
-      - name: Build EDoc
-        run: make docs
+      - name: Generate docs by ExDoc
+        run: rebar3 ex_doc
```
cw789, line 39: The run command could have stayed the same, now as I also changed the Makefile.

mworrell:
@cw789 Thanks!

cw789:
Thank you too. Hope it doesn't bring any burden to the one who do releases.

PR #309, opened 2022-02-12
Use OTP logger directly; drop `hut` dependency

seriyps:
Since we are already OTP-21+, there is no real reason to rely on `hut` for logger abstraction.

It replaces #267 since other PR seems stuck.

This PR also adds `domain` metadata to each log gen_smtp generates so it's easier to filter logs generated by `gen_smtp` using [logger_filters:domain](https://www.erlang.org/doc/man/logger_filters.html#domain-2).

Most of the changes were made by just running `sed` commands like this: `sed -i 's/\?log(warning, /\?LOG_WARNING(/' src/*.erl`


cw789 (commented):
Sanity checked the diff. Looks good as far as I could say.
Thanks. 💛

mworrell (approved):
Good to use OTP logger directly.

This is good for now, we can change to more structured logging at some later time.

Issue #310, opened 2022-02-15
Header param encoding


juhlig:
https://github.com/gen-smtp/gen_smtp/blob/c6f25a758d60da9788bf5ddf73e985dacb28f74b/src/mimemail.erl#L924-L929

I think this is wrong (though most Mail clients seem to accept it): Parameters should be encoded according to [RFC 2231](https://datatracker.ietf.org/doc/html/rfc2231), not RFC 2047.

seriyps:
Oh, interesting. This change have been merged quite recently https://github.com/gen-smtp/gen_smtp/pull/235. But I think it would worth to add support for RFC-2231 (in parser as well; including multiline params)

Maria-12648430:
#311 is about the encoding of parameter values. I'll see what I can do about decoding in a later PR (probably quite tricky, it looks like many clients do it wrong, so we need to be able to parse both RFC2231 _and_ RFC2047), but I think first this PR should be carefully reviewed.

PR #311, opened 2022-02-21
RFC2231 header parameter-value encoding

Maria-12648430:
As pointed out in #310, parameters are supposed to be encoded according to RFC2231, not RFC2047.

The value is first analyzed to decide which representation can/must be used, based on the presence of CTLs, spaces, tspecials, and UTF-8 multibyte characters. Then it is encoded in the selected representation. If the parameter lines become too long (>76 bytes), they are split up according to the continuation scheme described in RFC2231.

In hindsight, getting this right was surprisingly difficult, and therefore a careful review seems to be called for.

### DIFF
--- a/src/mimemail.erl
+++ b/src/mimemail.erl
@@ -916,28 +916,190 @@ has_lines_over_998(Bin, {FoundAt, 2}, _, Pattern) ->
         Bin, binary:match(Bin, Pattern, [{scope, {NewOffset, Len}}]), NewOffset, Pattern
     ).
 
+-spec encode_parameters([{Name :: binary(), Value :: binary()}]) -> [Parameter :: binary()].
 encode_parameters([[]]) ->
     [];
 encode_parameters(Parameters) ->
-    [encode_parameter(Parameter) || Parameter <- Parameters].
+    lists:foldr(
+        fun({Name, Value}, Acc) ->
+            {Method, EncLen} = decide_param_encoding_method(Value),
+            EncParams = encode_parameter(Method, Name, Value, EncLen),
+            EncParams ++ Acc
+        end,
+        [],
+        Parameters
+    ).
 
-encode_parameter({X, Y}) ->
-    YEnc = rfc2047_utf8_encode(Y, byte_size(X) + 3, <<"\t">>),
-    case escape_tspecial(YEnc, false, <<>>) of
-        {true, Special} -> [X, $=, $", Special, $"];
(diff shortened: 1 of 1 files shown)


seriyps:
Was surprized it did not trigger proper test failure, but it seems we don't generate long or non-ascii header parameter values.

https://github.com/gen-smtp/gen_smtp/blob/c6f25a758d60da9788bf5ddf73e985dacb28f74b/test/prop_mimemail.erl#L396

Might be nice to update when parser part is also implemented.

I'll have a deeper look bit later today, thak you!

Maria-12648430:
Last commit takes care of an edge case: invalid UTF-8 sequences were not encoded but left as-is.
Also added specs.

mworrell:
@seriyps Ok to merge this?

seriyps (approved):

seriyps:
Great work! Sorry for the delay, it was due to UA-RU war.

Maria-12648430:
@seriyps thanks 😄 TBH, I totally forgot about this one 😅

src/mimemail.erl
```diff
@@ -916,28 +916,195 @@ has_lines_over_998(Bin, {FoundAt, 2}, _, Pattern) ->
...
+    C =:= $];
+    C =:= $?;
+    C =:= $=;
+    C =:= $\s;
+    C =:= $*;
+    C =:= $';
+    C =:= $%
+->
+    BinC = <<C/utf8>>,
+    EncSize = 3 * byte_size(BinC),
```
mworrell, line 1044: Just wondering. Given the allowed values in the when-clause, isn't the byte size of BinC always 1?

mworrell (approved):

src/mimemail.erl
Maria-12648430: Hm, you're right, and also the subsequent handling of the character is "off". I mean, it does the right thing, but overly complicated. I'm not quite sure what I was thinking back then I have to admit, given that I did that almost a year ago :sweat_smile: I probably restructured things at some point and forgot to pull it all straight. I'll look into it again, but probably not today. So don't merge this yet :wink:

Maria-12648430: Done :)

mworrell: That looks a lot simpler :-)

mworrell:
@seriyps I am ok with this being merged.

PR #313, opened 2022-04-15
Remove legacy OTP support

cw789:
As OTP < 21 is not supported anymore we can rely on `ssl:handshake/1` & `ssl:handshake/3`.

### DIFF
--- a/src/smtp_util.erl
+++ b/src/smtp_util.erl
@@ -88,10 +88,8 @@ compute_cram_digest(Key, Data) ->
     Bin = hmac_md5(Key, Data),
     list_to_binary([io_lib:format("~2.16.0b", [X]) || <<X>> <= Bin]).
 
--ifdef(OTP_RELEASE).
 -if(?OTP_RELEASE >= 23).
 -define(CRYPTO_MAC, true).
--endif.
 -endif.
 
 -ifdef(CRYPTO_MAC).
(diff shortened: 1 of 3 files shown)


src/smtp_util.erl
```diff
@@ -88,11 +88,9 @@ compute_cram_digest(Key, Data) ->
     Bin = hmac_md5(Key, Data),
     list_to_binary([io_lib:format("~2.16.0b", [X]) || <<X>> <= Bin]).
 
--ifdef(OTP_RELEASE).
```
cw789, line 91: I'm not sure about this construct. But it seems to me it could be fine this way also.
Take with caution. I do not speak Erlang.

src/smtp_util.erl
```diff
@@ -88,11 +88,9 @@ compute_cram_digest(Key, Data) ->
...
 -if(?OTP_RELEASE >= 23).
 -define(CRYPTO_MAC, true).
 -endif.
--endif.
```
cw789, line 94: I'm not sure about this construct. But it seems to me it could be fine this way also.
Take with caution. I do not speak Erlang.

src/smtp_util.erl
```diff
@@ -88,11 +88,9 @@ compute_cram_digest(Key, Data) ->
...
 
 -ifdef(CRYPTO_MAC).
```
seriyps, line 95: I think you can just do this and remove all the 5 lines above
```suggestion
-if(?OTP_RELEASE >= 23).
```

seriyps (approved):

PR #321, opened 2022-09-03
Add LMTP support to the gen_smtp_client

pablowillians:
Hey team!

First of all, I would like to thank you guys for this exceptional work!
Second, I am proposing supporting LMTP in delivering messages.

It would be really useful in cases when it's needed to deliver messages to an LMTP server (ie, Dovecot)

Usage example:

```erlang
Socket = gen_smtp_client:open([{relay, "local-dovecot"}, {protocol, lmtp}])
gen_smtp_client:deliver(Socket, {"whatever@test.com", ["pablo@example.com"],
 "Subject: testing\r\nFrom: Andrew Thompson <andrew@hijacked.us>\r\nTo: Some Dude <foo@bar.com>\r\n\r\nThis is the email body"}).
```

I would love to receive suggestions to improve it. This is my first time writing Erlang code.

Thank you

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,6 +86,7 @@ The `send` method variants `send/2, send/3, send_blocking/2` take an `Options` a
   * **tls_options** used in `ssl:connect`, More info at http://erlang.org/doc/man/ssl.html . Defaults to `[{versions , ['tlsv1', 'tlsv1.1', 'tlsv1.2']}]`. This is merged with options listed at: https://github.com/gen-smtp/gen_smtp/blob/master/src/smtp_socket.erl#L46 . Any options not present in this list will be ignored.
   * **hostname** the hostname to be used by the smtp relay. Defaults to: `smtp_util:guess_FQDN()`. The hostname on your computer might not be correct, so set this to a valid value.
   * **retries** how many retries per smtp host on temporary failure. Defaults to 1, which means it will retry once if there is a failure.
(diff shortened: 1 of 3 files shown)


pablowillians:
@mworrell The workflow found an issue with formatting, so I updated the PR. Could you please approve the workflows to run again?

pablowillians:
Thanks, @mworrell ! I've made changes to the `rebar.config`'s file as well, since it was throwing a badmap error when trying to generate docs

mworrell:
So the basic difference is that we have to say LHLO? And all else is the same?

pablowillians:
> So the basic difference is that we have to say LHLO? And all else is the same?

Actually, it has one more difference:

> The change is that after the final ".", the server returns one reply for each previously successful RCPT command in the mail transaction

So I added [some tests](https://github.com/gen-smtp/gen_smtp/pull/321/files#diff-cfcb1b01c401c72da3137fa3166ce2ae9ea68669b62cb00f8b4f53a84d3f4f69R1199) similar to the [example from the LMTP documentation](https://datatracker.ietf.org/doc/html/rfc2033#section-4.1:~:text=EXAMPLE%3A%0A%0A%20%20%20S%3A%20220,edu%20closing%20connection), to reflect that behavior.

mworrell:
Looks good!  I like how you also added tests and updated the docs.

Thank you!

PR #325, opened 2023-04-18
Fix OTP 26 compilation failure

wingyplus:
Fix src/smtp_server_example.erl typespec in `handle_error` to use `#state{}` instead of `State`.

PR #326, opened 2023-04-18
Add OTP 25 and 26-rc3

wingyplus:
I just see from #325 is missing OTP 25 and 26 in the CI. So I add it to the workflow matrix.

For OTP 26, it has only rc versions so I add the latest rc (it's rc3 at the moment) to the workflow.


wingyplus:
Remove OTP 26 as per disscussed in #325

seriyps (approved):

mworrell:
Thanks!

PR #331, opened 2023-09-07
Add `auth_types` option, to avoid trying all auth types the server supports

justindotpub:
Currently gen_smtp will attempt to auth using every auth type supported by the server, in order. I want to use XOAUTH2 without waiting first for failed auth LOGIN attempts with username and password.

With the changes in this PR, you can pass a `auth_types` option that is a list of auth types you want to use, as opposed to iterating through all supported auth types.

I don't write much Erlang and welcome any feedback. Thank you!


justindotpub:
I ended up implementing https://github.com/swoosh/swoosh/pull/815 as an alternative to switching to smtp with oauth2 authentication, so I don't need this change any more.  If it isn't deemed valuable, feel free to close. 📕 Thanks. 🙏

PR #335, opened 2024-02-12
Document sockopts

TheOneric:
`tls_options` seemingly being ignored confused me for quiete a bit until I went through the code and found direct SSL connections actually need a different, not-yet-documented option.

To save others the trouble, here is a doc update :)


mworrell:
Thank you!!!

PR #337, opened 2024-02-27
Parse proxy protocol 

silviucpp:
In case you want to load balance the servers with an HAPROXY and you need to properly detect the peer ip you need to use the PROXY PROTOCOL: http://www.haproxy.org/download/1.8/doc/proxy-protocol.txt

### DIFF
--- a/src/gen_smtp_server_session.erl
+++ b/src/gen_smtp_server_session.erl
@@ -198,11 +198,22 @@ ranch_init({Ref, Transport, {Callback, Opts}}) ->
 -spec init(Args :: list()) -> {'ok', #state{}, ?TIMEOUT} | {'stop', any()} | 'ignore'.
 init([Ref, Transport, Socket, Module, Options]) ->
     Protocol = proplists:get_value(protocol, Options, smtp),
-    PeerName =
-        case Transport:peername(Socket) of
-            {ok, {IPaddr, _Port}} -> IPaddr;
-            {error, _} -> error
-        end,
+
+    PeerName = case proplists:get_value(proxy_protocol, Options, false) of
+        false ->
+            case Transport:peername(Socket) of
+                {ok, {IPaddr, _Port}} -> IPaddr;
+                {error, _} -> error
+            end;
+        _ ->
+            case read_proxy_protocol(Transport, Socket, <<>>) of
+                {ok, IpAddr} ->
+                    IpAddr;
+                _ ->
+                    error
+            end
+    end,
+
     case
         PeerName =/= error andalso
             Module:init(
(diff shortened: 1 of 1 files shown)


mworrell:
I guess it is a bit tricky to add a test for this?
(Considering you need some proxy to talk to...)

silviucpp:
@ mworrell  the fix is not proper. I will adjust it in the next days.

silviucpp:
I made the implementation much easy by using `{packet, line}`. Previous one was wrong because I was reading from socket also additional data coming after the proxy protocol.

silviucpp:
can be tested with haproxy basic config:

```
listen smtp-server

        bind :5878
        mode tcp
        option  tcplog
        balance roundrobin

        server smtp_tcp_s01 127.0.0.1:2525 check port 2525 send-proxy inter 4s rise 3 fall 2

```

src/gen_smtp_server_session.erl
```diff
@@ -1405,6 +1416,32 @@ report_recipient(multiple, [{ResponseType, Value} | Rest], State) ->
...
+% http://www.haproxy.org/download/1.8/doc/proxy-protocol.txt
+
+read_proxy_protocol(Transport, Socket) ->
+    ok = Transport:setopts(Socket, [{packet, line}, binary]),
+    case Transport:recv(Socket, 0, 5000) of
+        {ok, Data} ->
+            case Data of
+                <<"PROXY ", _/binary>> ->
+                    case binary:split(Data, <<" ">>, [global]) of
+                        [<<"PROXY">>, InetFamily, SrcIp, _DstIp, _SrcPort, _DstPort] when InetFamily == <<"TCP4">> orelse InetFamily == <<"TCP6">> ->
```
Vagabond, line 1429: You should probably use the right, address specific parser for v4 or v6, so if the proxy sends TCP4 with a v6 address, things fail.

src/gen_smtp_server_session.erl
```diff
@@ -1405,6 +1416,32 @@ report_recipient(multiple, [{ResponseType, Value} | Rest], State) ->
...
     report_recipient(multiple, Rest, State).
 
+% parse proxy protocol
+% http://www.haproxy.org/download/1.8/doc/proxy-protocol.txt
+
+read_proxy_protocol(Transport, Socket) ->
+    ok = Transport:setopts(Socket, [{packet, line}, binary]),
+    case Transport:recv(Socket, 0, 5000) of
+        {ok, Data} ->
+            case Data of
```
Vagabond, line 1426: There's also a binary form of this protocol (version 2), were you planning on supporting this as well?

src/gen_smtp_server_session.erl
silviucpp: Huh... can you be more explicit ?

src/gen_smtp_server_session.erl
silviucpp: We can support if you think it makes sense. Idea is that you have control in what version you want to use:

`send-proxy` is v1 (current implementation) 
`send-proxy-v2` - is v2

PR #338, opened 2024-04-08
smtp_server_example: fix DATA response

ts-klassen:
`["queued as ~s", Reference]` becomes like:

```
250 queued as ~sd98ae19ee87f0741ac9ba90d7046f0c5
```

I've fixed it to `["queued as ", Reference]` so that it looks like:

```
250 queued as d98ae19ee87f0741ac9ba90d7046f0c5
```


mworrell:
@ts-klassen Thank you!

PR #340, opened 2024-05-06
add tls_options to smtp:socket:connect/5 (erlang 26 fix)

sedinin:
One can provide e.g.
`{tls_options, [{verify, verify_none}]}`
to configuration to avoid problem introduced in erlang 26 (default `verify` is `verify_peer` now).

### DIFF
--- a/src/gen_smtp_client.erl
+++ b/src/gen_smtp_client.erl
@@ -850,13 +850,14 @@ connect(Host, Options) ->
             undefined -> [];
             Other -> Other
         end,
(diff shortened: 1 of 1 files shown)


Vagabond:
Can you not just add the tls options to sockoptions manually?

sedinin:
What do you mean by "manually"? I do it almost the same way its done in do_STARTTLS/2, 
by retrieving from configuration.

mworrell:
@seriyps Could you have a look at this?

PR #344, opened 2024-11-07
Allow sender to immediately hangup after QUIT message

mworrell:
This fixes a problem where the gen_smtp_server crashes if the client sends a QUIT message and immediately hangs up before the "221 Bye" can be send.  This is done by ignoring any send errors.

In the client:  be a bit more subtle when receiving funky multiline replies. Do not immediately crash but just throw an "unexpected_response" exception.

Use OTP 25 for docs build, as ex_doc requires it.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -17,7 +17,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
-        otp: ["24"] # https://www.erlang.org/downloads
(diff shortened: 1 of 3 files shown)


mworrell:
@arjan Could you have a look at this pull request?

arjan:
Looks good!

arjan (approved):

mworrell:
@arjan Shall we make a new release?  There are quite some changes since the 1.2.0.  (I don't have access to the Hex package, so can't release...)

PR #349, opened 2025-02-13
RFC-3030 (CHUNKING and BINARYMIME)

perfaram:
This PR implements [RFC 3030](https://datatracker.ietf.org/doc/html/rfc3030), which defines the CHUNKING and BINARYMIME extensions. 

This implementation is rather conservative in the changes it makes to gen_smtp, in particular regarding the API. 

On the server side, **by default**: 
* CHUNKING is enabled, as it is possible to fully abstract away how a message was delivered, from the PoV of a user of this library;
* BINARYMIME is disabled (unless activated in the `handle_EHLO` callback by setting it to `true`) because such messages always bypass the newline fixing and may not be UTF8, contrary to what one might reasonably expect when implementing `handle_DATA`. It is not currently possible to know whether a message was sent with this body type, should this be implemented? Feedback welcome here.

On the client side, **by default**: 
* CHUNKING (i.e., the `BDAT` command) will not be used, even if possible, **for now** (I'd like to have feedback first, but otherwise it should not be a problem to use it by default whenever possible);
* BINARYMIME will not be used, unless enabled through the options (so, session-wide -- I think it might make sense to allow setting this per-email? Feedback welcome here).

### DIFF
--- a/src/gen_smtp_client.erl
+++ b/src/gen_smtp_client.erl
@@ -38,7 +38,11 @@
     {retries, 1},
     {on_transaction_error, quit},
     % smtp, lmtp
-    {protocol, smtp}
+    {protocol, smtp},
+    % default (max) chunk size for the RFC-3030 CHUNKING extension
+    {chunk_size, 256 * 1024},
+    % default to not use binarymime
+    {use_binarymime, false}
 ]).
 
 -define(AUTH_PREFERENCE, [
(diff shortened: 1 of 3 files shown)


mworrell:
@seriyps On first glance this looks good to me, and all tests are passing.  Can you add a review?

PR #358, opened 2026-06-04
Add eiconv as a dependency

Munksgaard:
Previously, eiconv was only added in the dialyzer and test profiles. But iconv is used in e.g. `mimemail.convert/3`.

### DIFF
--- a/rebar.config
+++ b/rebar.config
@@ -44,7 +44,8 @@
 ]}.
 
 {deps, [
-    {ranch, ">= 1.8.0"}
+    {ranch, ">= 1.8.0"},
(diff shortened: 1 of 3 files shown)


Vagabond:
The idea was that eiconv was not *required*

Munksgaard:
> The idea was that eiconv was not _required_

But [it is required by `mimemail:convert/3` right now](https://github.com/gen-smtp/gen_smtp/blob/master/src/mimemail.erl#L339). I don't know how one usually handle those things in Erlang, but in Elixir you'd add a compile-time condition around `mimemail:convert/3` that checks whether `iconv:convert/3` is available.

PR #359, opened 2026-07-30
Add `none` as an encoding option in spec for `decode`

Munksgaard:
Previously, the type for `decode` didn't mention `none` as a valid value for the `encoding` key, even though the code supports `none`. This created dialyzer errors in dependent applications.

This PR adds `none` to the type by adding a new intermediary type called `encoding`.


Vagabond:
This generally looks ok to me

